PR #4, opened 2024-05-12
Set $MESA_LOADER_DRIVER_OVERRIDE to "asahi" to make Windows visible.

SO5005:


teohhanhui:
I had a commit where we were calling `LANG=C eglinfo` and just looking for `Driver name: asahi`. Is that okay? 

It'd be very complex to try to call EGL APIs directly.

slp:
> I had a commit where we were calling `LANG=C eglinfo` and just looking for `EGL driver name: asahi`. Is that okay?
> 
> [slp/krun.legacy#4 (comment)](https://github.com/slp/krun.legacy/pull/4#discussion_r1590459236)
> 
> It'd be very complex to try to call EGL APIs directly. But sure, we could do it if you don't mind more code... (Will just port it from here: https://github.com/mpv-player/mpv/blob/d347ee0e916de83e74fc4dca1d35300307ce3f6b/video/out/opengl/egl_helpers.c#L313)

I think it's better to just check `/proc/device-tree/model` or `/proc/device-tree/compatible`. We can take advantage that GPUs are embedded in this computers so we can safely assume that the asahi mesa driver must be used on them.

@SO5005 do you want to give this approach a try on this PR?

crates/krun/src/bin/krun.rs
```diff
@@ -1,6 +1,7 @@
 use std::{
     env::{self, VarError},
     ffi::{c_char, CString},
+    fs::read_to_string,
```
teohhanhui, line 4: minor style nit:

Only import `fs` here, and do `fs::read_to_string` at the call site below.

Same as the style in the example:
https://doc.rust-lang.org/stable/std/fs/fn.read_to_string.html

crates/krun/src/bin/krun.rs
```diff
@@ -206,10 +207,25 @@ fn main() -> Result<()> {
...
+        "t8103", "t6000", "t6001", "t6002", "t8112", "t6020", "t6021", "t6022", "t8121", "t6030",
+        "t6031", "t6034",
+    ];
     for key in WELL_KNOWN_ENV_VARS {
         let value = match env::var(key) {
             Ok(value) => value,
             Err(VarError::NotPresent) => {
+                if key == "MESA_LOADER_DRIVER_OVERRIDE" {
+                    for compatible in
+                        read_to_string("/proc/device-tree/compatible")?.split(&[',', '\0'][..])
```
teohhanhui, line 222: I think we must not error here on devices that don't have `/proc/device-tree`. So just do `.ok()`

teohhanhui: Or rather, we can still propage other `io::Error` that are not `ErrorKind::NotFound`:
https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.NotFound

WDYT?

crates/krun/src/bin/krun.rs
teohhanhui: (This is mainly because there are both `fs::read_to_string` and `io::read_to_string` in std... So it would be ambiguous to the reader at a glance.)

crates/krun/src/bin/krun.rs
```diff
@@ -206,10 +207,25 @@ fn main() -> Result<()> {
         "PATH",                        // needed by `krun-guest` program
         "RUST_LOG",
     ];
+
+    // https://github.com/AsahiLinux/docs/wiki/Devices
+    const ASAHI_DEVS: [&str; 12] = [
+        "t8103", "t6000", "t6001", "t6002", "t8112", "t6020", "t6021", "t6022", "t8121", "t6030",
```
teohhanhui, line 213: Typo

```suggestion
        "t8103", "t6000", "t6001", "t6002", "t8112", "t6020", "t6021", "t6022", "t8122", "t6030",
```

crates/krun/src/bin/krun.rs
```diff
@@ -206,10 +207,25 @@ fn main() -> Result<()> {
         "PATH",                        // needed by `krun-guest` program
         "RUST_LOG",
     ];
+
+    // https://github.com/AsahiLinux/docs/wiki/Devices
+    const ASAHI_DEVS: [&str; 12] = [
```
teohhanhui, line 212:
```suggestion
    const ASAHI_SOC_IDS: [&str; 12] = [
```

crates/krun/src/bin/krun.rs
```diff
@@ -206,10 +207,25 @@ fn main() -> Result<()> {
...
+    ];
     for key in WELL_KNOWN_ENV_VARS {
         let value = match env::var(key) {
             Ok(value) => value,
             Err(VarError::NotPresent) => {
+                if key == "MESA_LOADER_DRIVER_OVERRIDE" {
+                    for compatible in
+                        read_to_string("/proc/device-tree/compatible")?.split(&[',', '\0'][..])
+                    {
+                        if ASAHI_DEVS.iter().any(|&s| s == compatible) {
```
teohhanhui, line 224: I don't think this would match? For example, it's `t8112apple` on my machine.

```
teohhanhui@han-macbook-air:~$ cat /proc/device-tree/compatible
apple,j413apple,t8112apple,arm-platform
```

teohhanhui: Oh...

```
teohhanhui@han-macbook-air:~/Projects/FEX-Emu/RootFS$ od -c /proc/device-tree/compatible
0000000   a   p   p   l   e   ,   j   4   1   3  \0   a   p   p   l   e
0000020   ,   t   8   1   1   2  \0   a   p   p   l   e   ,   a   r   m
0000040   -   p   l   a   t   f   o   r   m  \0
0000052
```

But I think that means we should match on `t8112\0apple` instead?

SO5005: Having the `'\0'` in the middle of the string looks kind strange.

teohhanhui: I found it confusing at first, but it's a genius thing for them to do... The users who simply `cat /proc/device-tree/compatible` would see it as `t8112apple`. But there's a hidden NUL byte for programs to parse, in case they want to separate the different components.

SO5005: It took me way too long to find out that the NUL was in the middle and not at the end.

crates/krun/src/bin/krun.rs
```diff
@@ -206,10 +207,26 @@ fn main() -> Result<()> {
...
+        "t6031", "t6034",
+    ];
     for key in WELL_KNOWN_ENV_VARS {
         let value = match env::var(key) {
             Ok(value) => value,
             Err(VarError::NotPresent) => {
+                if key == "MESA_LOADER_DRIVER_OVERRIDE" {
+                    for compatible in fs::read_to_string("/proc/device-tree/compatible")
+                        .unwrap_or("".to_string())
+                        .split(&[',', '\0'][..])
```
teohhanhui, line 223:
```suggestion
                    match fs::read_to_string("/proc/device-tree/compatible") {
                        Ok(compatible) => {
                            // ...
                        },
                        Err(err) => if err.kind() == ErrorKind::NotFound {
                            continue;
                        },
                        Err(err) => Err(err).context("Failed to read `/proc/device-tree/compatible`")?,
                    }
```

crates/krun/src/bin/krun.rs
```diff
@@ -206,10 +207,32 @@ fn main() -> Result<()> {
...
     for key in WELL_KNOWN_ENV_VARS {
         let value = match env::var(key) {
             Ok(value) => value,
             Err(VarError::NotPresent) => {
+                if key == "MESA_LOADER_DRIVER_OVERRIDE" {
+                    match fs::read_to_string("/proc/device-tree/compatible") {
+                        Ok(compatible) => {
+                            // ...
+                        },
+                        Err(err) => if err.kind() == ErrorKind::NotFound {
```
teohhanhui, line 225: Oops. I had a typo here.
```suggestion
                        Err(err) if err.kind() == ErrorKind::NotFound => {
```

crates/krun/src/bin/krun.rs
teohhanhui: You know what... I should have read the docs. I was completely mistaken. :laughing:

https://docs.kernel.org/devicetree/usage-model.html

So on my machine it's actually:

"apple,j413", "apple,t8112", "apple,arm-platform"

The NUL byte (`\0`) is the separator.

SO5005: Ok, that makes more sense.

SO5005: Fixed it.

crates/krun/src/bin/krun.rs
```diff
@@ -206,10 +208,44 @@ fn main() -> Result<()> {
         "PATH",                        // needed by `krun-guest` program
         "RUST_LOG",
     ];
+
+    // https://github.com/AsahiLinux/docs/wiki/Devices
+    const ASAHI_SOC_DEV_IDS: [&str; 12] = [
```
teohhanhui, line 213:
```suggestion
    const ASAHI_SOC_COMPAT_IDS: [&str; 12] = [
```

My bad. dev_id / device id has a specific meaning, which would be confusing in this case.

slp (approved):
LGTM, thanks!

PR #7, opened 2024-05-17
CI: add code quality check

slp:
Add a code quality check that runs both rustfmt and clippy on the PR to enforce both consistency and tidiness.


slp:
Turning it into a draft until we have a 1.9.0 packaged in Fedora so we can use for linking when running the tests.

teohhanhui:
It was superseded by #67

PR #11, opened 2024-05-26
Default guest memory to 80% of total RAM on host

teohhanhui:
From https://vt.social/@lina/112507831878162342

### DIFF
--- a/crates/krun/src/bin/krun.rs
+++ b/crates/krun/src/bin/krun.rs
@@ -1,4 +1,5 @@
 use std::{
+    cmp,
     env::{self, VarError},
     ffi::{c_char, CString},
     fs,
@@ -9,7 +10,8 @@ use std::{
 use anyhow::{anyhow, Context, Result};
 use krun::{
     cli_options::options,
-    net::{connect_to_passt, start_passt, NetMode},
+    net::{connect_to_passt, start_passt},
+    types::{MiB, NetMode},
 };
 use krun_sys::{
     krun_create_ctx, krun_set_exec, krun_set_gpu_options, krun_set_log_level, krun_set_passt_fd,
@@ -17,7 +19,7 @@ use krun_sys::{
     VIRGLRENDERER_THREAD_SYNC, VIRGLRENDERER_USE_ASYNC_FENCE_CB, VIRGLRENDERER_USE_EGL,
 };
 use log::debug;
-use nix::unistd::User;
+use nix::{sys::sysinfo::sysinfo, unistd::User};
 use rustix::{
     io::Errno,
     process::{geteuid, getgid, getrlimit, getuid, setrlimit, Resource},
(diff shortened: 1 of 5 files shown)


crates/krun/src/bin/krun.rs
```diff
@@ -58,10 +60,17 @@ fn main() -> Result<()> {
...
     {
-        // Configure the number of vCPUs (4) and the amount of RAM (4096 MiB).
+        let ram_mib = if let Some(ram_mib) = options.mem {
+            ram_mib
+        } else {
+            let sysinfo = sysinfo().context("Failed to get system information")?;
+            let ram_total = sysinfo.ram_total() / 1024 / 1024;
+            cmp::min(MiB::from((ram_total as f64 * 0.8) as u32), MiB::from(16384))
+        };
+        // Configure the number of vCPUs (4) and the amount of RAM (max 16384 MiB).
```
teohhanhui, line 70: (From https://github.com/teohhanhui/krun/commit/ce29e32992389569b70076ccc791577b8d0cfe28#r142414630)

@asahilina
> Is there a reason for this limit? We have machines with 192GB of RAM, surely it makes sense for the VM to support more than 16G...

I have no idea why there's this limit in krunvm. Was going to ask @slp since he's the original author of that code according to git blame. Will remove once we confirm that it's not needed.

crates/krun/src/cli_options.rs
```diff
@@ -29,6 +30,22 @@ pub fn options() -> OptionParser<Options> {
...
+            immediately. Instead, it will be provided as the guest demands it,
+            and both the guest and libkrun (acting as the Virtual Machine
+            Monitor) will attempt to return as many pages as possible to the
+            host.
+    [default: 80% of total RAM]",
+        )
+        .argument("MEM")
+        .guard(
+            |&mem| mem <= MiB::from(16384),
+            "the maximum amount of RAM supported is 16384 MiB",
```
teohhanhui, line 46: (Moved from https://github.com/teohhanhui/krun/commit/ce29e32992389569b70076ccc791577b8d0cfe28#r142414630)

@asahilina
> Is there a reason for this limit? We have machines with 192GB of RAM, surely it makes sense for the VM to support more than 16G...

I have no idea why there's this limit in krunvm. Was going to ask @slp since he's the original author of that code according to git blame. Will remove once we confirm that it's not needed.

crates/krun/src/bin/krun.rs
teohhanhui, line 70: (Moved from https://github.com/teohhanhui/krun/commit/ce29e32992389569b70076ccc791577b8d0cfe28#r142414585)

@asahilina:
> Aside, but for vCPUs, what I would do is enumerate the system CPUs, filter out the P-cores (the cores with the highest capacity), use the count as the vCPU count, and then automatically pin the krun VM to those CPUs. That sidesteps the big.LITTLE VM scheduling issue (which is a much bigger problem) ^^

crates/krun/src/cli_options.rs
slp: Originally, libkrunfw's kernel was limited to 8 vCPU and 16GB to reduce the kernel size, since it was intended to be used for container isolation exclusively. Since we've now incorporated other use cases, we should lift that limit. It needs to be done in libkrun and libkrunfw first, so it makes sense to keep that limit in krun in the meantime.

crates/krun/src/bin/krun.rs
slp: Sounds like a good strategy to me too. @teohhanhui do you want to extend this PR to cover that or should we leave that for later.

slp:
LGTM, thanks!

slp (approved):

PR #12, opened 2024-05-26
Raise RLIMIT_NOFILE in krun-guest

teohhanhui:
This is required for wine's esync to work.

### DIFF
--- a/crates/krun-guest/Cargo.toml
+++ b/crates/krun-guest/Cargo.toml
@@ -14,7 +14,7 @@ bpaf = { workspace = true, features = [] }
 env_logger = { workspace = true, features = ["auto-color", "humantime", "unstable-kv"] }
 log = { workspace = true, features = ["kv"] }
 nix = { workspace = true, features = ["user"] }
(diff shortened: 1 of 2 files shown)


slp (approved):
LGTM, thanks.

teohhanhui:
Oops... This did not solve the problem. We actually need to increase the hard limit here, not the soft limit.

@slp Do you think this is something that would be appropriate to do in `krun`? I believe the `RLIMIT_NOFILE` here is just the kernel default, since nothing else (usually systemd, PAM) is changing it...

teohhanhui:
The `systemd` default for the hard limit is `524288`:

https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html#DefaultLimitCPU=

As can also be seen in `/usr/lib/systemd/system.conf` on a Fedora system.

teohhanhui:
> In other words, if we need NOFILE to be `524288` (or whatever we need for wine esync), let's just set both hard and soft limits to that in `krun-guest`.

Included in #28

PR #23, opened 2024-06-01
Introduce a client-server model

slp:
With the introduction of 47fdd45a (Default guest memory to 80% of total RAM on host), we're targeting a use case in which we expect to be running a single, large VM.

To accommodate this use case, we're switching here to a client/server model. The first time krun is launched, in addition to running the command, it starts a small server in the VM listening on a TCP port. On the host side, it also creates a lock file, writes the listening port number to it, and calls flock() on it to prevent other instances from running in parallel.

When it's executed again, krun detects there's another instance running, and contacts the server in the VM to request the launch of the command. The stdout and stderr of the command are redirected to $XDG_RUNTIME_DIR/krun-$COMMAND-$TIMESTAMP.[stdout|stderr]

### DIFF
--- a/crates/krun-server/Cargo.toml
+++ b/crates/krun-server/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "krun"
+name = "krun-server"
 version = { workspace = true }
 authors = { workspace = true }
 edition = { workspace = true }
@@ -12,10 +12,8 @@ license = { workspace = true }
 anyhow = { workspace = true, features = ["std"] }
 bpaf = { workspace = true, features = [] }
 env_logger = { workspace = true, features = ["auto-color", "humantime", "unstable-kv"] }
-krun-sys = { workspace = true, features = [] }
 log = { workspace = true, features = ["kv"] }
-nix = { workspace = true, features = ["user"] }
-rustix = { workspace = true, features = ["process", "std", "use-libc-auxv"] }
+nix = { workspace = true, features = ["socket"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 utils = { workspace = true, features = [] }
--- a/crates/krun-server/src/bin/krun-server.rs
+++ b/crates/krun-server/src/bin/krun-server.rs
@@ -1,0 +1,33 @@
+use std::net::TcpListener;
+use std::os::fd::AsRawFd;
+use std::process::Command;
+
+use anyhow::Result;
+use krun_server::cli_options::options;
+use krun_server::server::start_server;
+use nix::sys::socket::{shutdown, Shutdown};
+
+fn main() -> Result<()> {
+    env_logger::init();
+
+    let options = options().run();
+
+    let listener = TcpListener::bind(format!("127.0.0.1:{}", &options.server_port))?;
+    let listener_fd = listener.as_raw_fd();
+
+    let server_thread = start_server(listener);
+
+    let status = Command::new(&options.command)
+        .args(options.command_args)
+        .status();
+    if let Err(err) = status {
+        println!("Error executing command {}: {}", options.command, err);
+    }
+
+    shutdown(listener_fd, Shutdown::Both)?;
+    if let Err(err) = server_thread.join() {
+        println!("Error waiting for server thread termination: {err:?}");
+    }
+
+    Ok(())
+}
--- a/crates/krun-server/src/cli_options.rs
+++ b/crates/krun-server/src/cli_options.rs
@@ -1,0 +1,40 @@
+use bpaf::{any, construct, env, positional, OptionParser, Parser};
+
+#[derive(Clone, Debug)]
+pub struct Options {
+    pub server_port: u32,
+    pub command: String,
+    pub command_args: Vec<String>,
+}
+
+pub fn options() -> OptionParser<Options> {
+    let server_port = env("KRUN_SERVER_PORT")
+        .short('p')
+        .help("TCP port to listen for command launch requests")
+        .argument("SERVER_PORT")
+        .fallback(3333)
+        .display_fallback();
+    let command = positional("COMMAND");
+    let command_args = any::<String, _, _>("COMMAND_ARGS", |arg| {
+        (!["--help", "-h"].contains(&&*arg)).then_some(arg)
+    })
+    .many();
+
+    construct!(Options {
+        server_port,
+        // positionals
+        command,
+        command_args,
+    })
+    .to_options()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn check_options() {
+        options().check_invariants(false)
+    }
+}
--- a/crates/krun/src/bin/krun.rs
+++ b/crates/krun/src/bin/krun.rs
@@ -12,6 +12,7 @@ use anyhow::{anyhow, Context, Result};
 use krun::{
     cli_options::options,
     cpu::{get_fallback_cores, get_performance_cores},
+    lock::{lock_or_connect, LockResult},
     net::{connect_to_passt, start_passt},
     types::{MiB, NetMode},
 };
@@ -27,7 +28,7 @@ use rustix::{
     io::Errno,
     process::{geteuid, getgid, getrlimit, getuid, sched_setaffinity, setrlimit, CpuSet, Resource},
 };
-use utils::env::find_in_path;
+use utils::env::find_krun_exec;
 
 fn main() -> Result<()> {
     env_logger::init();
@@ -38,6 +39,15 @@ fn main() -> Result<()> {
     }
 
     let options = options().fallback_to_usage().run();
+
+    let _lock_file = match lock_or_connect(&options)? {
+        LockResult::LaunchRequested => {
+            // There was a krun instance already running and we've requested it
+            // to launch the command successfully, so all the work is done.
+            return Ok(());
+        },
+        LockResult::LockAcquired(lock_file) => lock_file,
+    };
 
     {
         // Set the log level to "off".
@@ -136,7 +146,9 @@ fn main() -> Result<()> {
                 .context("Failed to connect to `passt`")?
                 .into()
         } else {
-            start_passt().context("Failed to start `passt`")?.into()
+            start_passt(options.server_port)
+                .context("Failed to start `passt`")?
+                .into()
         };
         // SAFETY: `passt_fd` is an `OwnedFd` and consumed to prevent closing on drop.
         // See https://doc.rust-lang.org/std/io/index.html#io-safety
@@ -196,28 +208,8 @@ fn main() -> Result<()> {
         }
     }
 
-    let krun_guest_path =
-        find_in_path("krun-guest").context("Failed to check existence of `krun-guest`")?;
-    let krun_guest_path = if let Some(krun_guest_path) = krun_guest_path {
-        krun_guest_path
-    } else {
-        let krun_path = env::current_exe().and_then(|p| p.canonicalize());
-        let krun_path = krun_path.context("Failed to get path of current running executable")?;
-        krun_path.with_file_name(format!(
-            "{}-guest",
-            krun_path
-                .file_name()
-                .expect("krun_path should end with a file name")
-                .to_str()
-                .context("Failed to process `krun` file name as it contains invalid UTF-8")?
-        ))
-    };
-    let krun_guest_path = CString::new(
-        krun_guest_path
-            .to_str()
-            .context("Failed to process `krun-guest` path as it contains invalid UTF-8")?,
-    )
-    .context("Failed to process `krun-guest` path as it contains NUL character")?;
+    let krun_guest_path = find_krun_exec("krun-guest")?;
+    let krun_server_path = find_krun_exec("krun-server")?;
 
     let mut krun_guest_args: Vec<CString> = vec![
         CString::new(username).expect("username should not contain NUL character"),
@@ -226,6 +218,8 @@ fn main() -> Result<()> {
         CString::new(format!("{}", getgid().as_raw()))
             .expect("gid should not contain NUL character"),
     ];
+
+    krun_guest_args.push(krun_server_path);
     krun_guest_args.push(
         CString::new(options.command)
             .context("Failed to process command as it contains NUL character")?,
@@ -236,6 +230,7 @@ fn main() -> Result<()> {
             .context("Failed to process command arg as it contains NUL character")?;
         krun_guest_args.push(s);
     }
+
     let krun_guest_args: Vec<*const c_char> = {
         const KRUN_GUEST_ARGS_FIXED: usize = 4;
         // SAFETY: All pointers must be stored in the same allocation.
@@ -316,6 +311,8 @@ fn main() -> Result<()> {
         })?;
         env.push(s);
     }
+
+    env.push(CString::new(format!("KRUN_SERVER_PORT={}", options.server_port)).unwrap());
 
     debug!(env:?; "env vars");
 
--- a/crates/utils/src/launch.rs
+++ b/crates/utils/src/launch.rs
@@ -1,0 +1,10 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Eq, PartialEq, Debug, Deserialize, Serialize)]
+pub struct Launch {
+    pub command: String,
+    pub args: Vec<String>,
+    pub envs: HashMap<String, String>,
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,6 +208,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "itoa"
+version = "1.0.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "49f1f14873335454500d59611f1cf4a4b0f786f9ac11f4312a78e4cf2566695b"
+
+[[package]]
 name = "krun"
 version = "0.1.0"
 dependencies = [
@@ -218,6 +224,8 @@ dependencies = [
  "log",
  "nix",
  "rustix",
+ "serde",
+ "serde_json",
  "utils",
 ]
 
@@ -232,6 +240,19 @@ dependencies = [
  "nix",
  "rustix",
  "tempfile",
+ "utils",
+]
+
+[[package]]
+name = "krun-server"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "bpaf",
+ "env_logger",
+ "log",
+ "serde",
+ "serde_json",
  "utils",
 ]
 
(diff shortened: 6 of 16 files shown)


teohhanhui:
Any reason for using a TCP port vs a Unix socket? Would vsock not work well for that?

slp:
> Any reason for using a TCP port vs a Unix socket? Would vsock not work well for that?

Right now libkrun only supports outgoing (VM -> host) connections with vsock. Implementing it the other way around is doable, but would be quite expensive to make it reliable, specially compared with simply using a TCP socket.

crates/krun-guest/src/bin/krun-guest.rs
```diff
@@ -37,14 +37,16 @@ fn main() -> Result<()> {
 
     setup_pulse_proxy(run_path)?;
 
+    let command = match options.command {
+        Some(c) => c,
+        None => "/usr/local/bin/krun-server".to_string(),
+    };
```
teohhanhui, line 43:
```suggestion
    let command = options.command.unwrap_or("/usr/local/bin/krun-server".to_owned());
```

crates/krun-guest/src/cli_options.rs
```diff
@@ -10,7 +10,7 @@ pub struct Options {
     pub username: String,
     pub uid: Uid,
     pub gid: Gid,
-    pub command: String,
+    pub command: Option<String>,
```
teohhanhui, line 13: I never realized this when I did the rewrite, but shouldn't we use `PathBuf` for command?

crates/krun-guest/src/cli_options.rs
```diff
@@ -26,7 +26,7 @@ pub fn options() -> OptionParser<Options> {
             .context("Failed to parse GID")
             .map(|gid| gid.into())
     });
-    let command = positional("COMMAND");
+    let command = positional("COMMAND").optional();
```
teohhanhui, line 29: Actually I don't think this can work correctly. What does an optional positional argument that's followed by non-optional one(s) even mean?

crates/krun-server/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+nix = { workspace = true, features = ["user"] }
+rustix = { workspace = true, features = ["fs", "mount", "std", "system", "use-libc-auxv"] }
+tempfile = { workspace = true, features = [] }
+utils = { workspace = true, features = [] }
+vsock = "0.5.0"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+
+[features]
+default = []
```
teohhanhui, line 25: Missing newline at end of file. I'd recommend setting up EditorConfig for your code editor to avoid this in the future...

crates/krun-server/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+license = { workspace = true }
+
+[dependencies]
+anyhow = { workspace = true, features = ["std"] }
+bpaf = { workspace = true, features = [] }
+env_logger = { workspace = true, features = ["auto-color", "humantime", "unstable-kv"] }
+log = { workspace = true, features = ["kv"] }
+nix = { workspace = true, features = ["user"] }
+rustix = { workspace = true, features = ["fs", "mount", "std", "system", "use-libc-auxv"] }
+tempfile = { workspace = true, features = [] }
```
teohhanhui, line 18: Unused dependencies.

```suggestion
```

crates/utils/src/launch.rs
```diff
@@ -0,0 +1,51 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Launch {
+    pub binary: String,
+    pub args: Vec<String>,
+    pub envs: HashMap<String, String>,
+}
+
+pub fn request_launch(
```
teohhanhui, line 17: I think this should live in the `krun` crate?

crates/krun-server/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+utils = { workspace = true, features = [] }
+vsock = "0.5.0"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
```
teohhanhui, line 22: Unused dependencies.

```suggestion
```

crates/krun-server/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[dependencies]
+anyhow = { workspace = true, features = ["std"] }
+bpaf = { workspace = true, features = [] }
+env_logger = { workspace = true, features = ["auto-color", "humantime", "unstable-kv"] }
+log = { workspace = true, features = ["kv"] }
+nix = { workspace = true, features = ["user"] }
+rustix = { workspace = true, features = ["fs", "mount", "std", "system", "use-libc-auxv"] }
+tempfile = { workspace = true, features = [] }
+utils = { workspace = true, features = [] }
+vsock = "0.5.0"
```
teohhanhui, line 20: Unused dependency.

```suggestion
```

crates/krun-server/Cargo.toml
teohhanhui, line 22: Wrongly sorted dependencies. Use `cargo add` in the future to avoid this problem.

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,31 @@
+use std::env;
+use std::process::Command;
+use std::thread;
```
teohhanhui, line 3: CS: merge imports

```suggestion
use std::{env, process::Command, thread};
```

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use anyhow::Result;
+use krun_server::cli_options::options;
+use krun_server::server::start_server;
```
teohhanhui, line 7: CS: merge imports

```suggestion
use krun_server::{cli_options::options, server::start_server};
```

teohhanhui: But yeah, we could change this if you'd prefer. The current Rust style guide says:

> Prefer to use multiple imports rather than a multi-line import.

https://github.com/rust-lang/rust/blob/9f2d0b3490729ed65da73f64b28d8d1f197af113/src/doc/style-guide/src/items.md#large-list-imports

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,31 @@
...
+
+fn main() -> Result<()> {
+    env_logger::init();
+
+    let options = options().run();
+
+    let server_port = env::var("KRUN_SERVER_PORT").unwrap_or("3333".to_string());
```
teohhanhui, line 14: This could be in `cli_options` (`--port` / `-p`?)

`bpaf` has support for parsing env var too: https://docs.rs/bpaf/latest/bpaf/fn.env.html

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,31 @@
...
+    let server_port: u32 = server_port.parse().unwrap_or(3333);
+
+    thread::spawn(move || {
+        if let Err(e) = start_server(server_port) {
+            println!("Error starting krun server: {e}")
+        }
```
teohhanhui, line 20:
```suggestion
        if let Err(err) = start_server(server_port) {
            println!("Failed to start krun server: {err?}")
        }
```

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,31 @@
...
+    });
+
+    let err = Command::new(&options.command)
+        .args(options.command_args)
+        .status();
+    if let Err(err) = err {
+        println!("Error executing command {}: {}", options.command, err);
+    }
+
+    std::process::exit(0);
```
teohhanhui, line 30: Do we not intend to keep the VM running in the background? To me that seems like it'd be much better for UX, in the sense that running a command via `krun` would be instant if the VM / `krun-server` is already running in the background.

teohhanhui: I'm imagining a UX like this:

```
krun-server start
krun ...
krun ...
krun-server stop
```

Which makes it kinda analogous to `podman start` / `podman stop`?

crates/krun-guest/src/bin/krun-guest.rs
slp: In which sense? As an option to specific the name of the `krun-server`?

crates/krun-guest/src/cli_options.rs
slp: That's a left over, I'll fix it.

crates/utils/src/launch.rs
slp: Hm... okay.

crates/krun-server/src/bin/krun-server.rs
slp: It's actually more convenient having it as an environment variable as it severs the double purpose of allowing us to know if krun is being run in the VM itself.

crates/krun-guest/src/bin/krun-guest.rs
teohhanhui: This is using `krun-server` as the default `command`, right? What I meant is that we could set this fallback value in `cli_options` instead of here.

crates/utils/src/launch.rs
teohhanhui: The way I see it, the `utils` crate should be for stuff that are common between the different crates / binaries. And any dependency added to the `utils` crate should ideally be optional, gated behind a feature flag.

crates/krun-server/src/bin/krun-server.rs
teohhanhui: Sure, but we don't have to do the parsing manually here if we could let `bpaf` do it for us? And it'd be more uniform when we add more options.

crates/krun-server/src/bin/krun-server.rs
slp: > Do we not intend to keep the VM running in the background? To me that seems like it'd be much better for UX, in the sense that running a command via krun would be instant if the VM / krun-server is already running in the background.

This is what happens with this PR. If krun is already running, launching the command feels instantaneous.

> I'm imagining a UX like this:
>
> krun-server start
> krun ...
> krun ...
> krun-server stop 
> Which makes it kinda analogous to podman start / podman stop?
>
> (krun can autostart krun-server if it's not already running? Socket activation?)

The UX implemented in this PR is better. It allows users to preserve the current workflow, just running a single command that will tear down the VM when finished (which is the spirit of microVMs anyways), but it also allows users to run multiple commands on the same VM seamlessly. It's even compatible with binfmt (I've tested it this weekend).

Starting the command will always feel instantaneous. The only reason it takes a bit now when VM is not running is due to `dhcpcd` doing stupid things, something I intend to fix in a future PR.

Installing krun as a systemd socket-activated service will imply an implicit, hidden lifetime for the VM, which can be confusing (why is krun still running if I've just closed the app?).

crates/krun-guest/src/cli_options.rs
slp: Yes, let's do that in a follow-up PR.

crates/krun-server/src/bin/krun-server.rs
slp: Let's make this homogeneous in a follow-up PR.

crates/krun-server/src/bin/krun-server.rs
slp: Ah, yes, I didn't get that on the first read... 😅 Will update the PR.

crates/krun/src/bin/krun.rs
```diff
@@ -37,7 +38,16 @@ fn main() -> Result<()> {
...
 
-    let options = options().fallback_to_usage().run();
+    let options = options().run();
+
+    let lock_file = lock_or_connect(&options)?;
+    if lock_file.is_none() {
+        // If lock_or_connect has returned without error and without
+        // a lock file, it means there was already a krun instance
+        // running and we've requested it to launch the command
+        // successfully, so all the work is done.
```
teohhanhui, line 48: We should use an `enum` to represent the different cases instead of bespoke logic... :see_no_evil:

crates/krun-guest/src/bin/krun-guest.rs
slp: This was a leftover from a previous approach. I've dropped it.

crates/krun-guest/src/cli_options.rs
slp: Done.

crates/krun-server/Cargo.toml
slp: Fixed.

crates/utils/src/launch.rs
slp: Done.

crates/krun-server/Cargo.toml
slp: Dropped.

crates/krun-server/Cargo.toml
slp: Moved them to workspace, since they're now used by krun-server and krun.

crates/krun-server/src/bin/krun-server.rs
slp: Done with bpaf, thanks!

crates/krun/src/bin/krun.rs
slp: Done.

crates/krun-server/src/bin/krun-server.rs
slp: Applied, thanks.

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,27 @@
...
+use anyhow::Result;
+use krun_server::cli_options::options;
+use krun_server::server::start_server;
+
+fn main() -> Result<()> {
+    env_logger::init();
+
+    let options = options().run();
+
+    thread::spawn(move || {
```
teohhanhui, line 13: I believe we might need a channel for graceful shutdown of the server thread... Otherwise no cleanup can happen and the TCP socket is not closed. Is that okay?

slp: The moment `options.command` exits, we exit ourselves and the thread and all its associated resources (both kernel and userspace) are released. We would need a sync mechanism if we were spawning a process (fork+execve), but with a thread we can avoid implementing that (in our case) unnecessary boilerplate.

teohhanhui: Oh, what I mean is to add a `channel`, so that we can send a signal to trigger a graceful shutdown in the server thread, and then we can `join` the server thread to wait for the server thread to fully shutdown before we return from main.

Related:
* https://doc.rust-lang.org/book/ch16-02-message-passing.html
* https://doc.rust-lang.org/book/ch20-03-graceful-shutdown-and-cleanup.html

teohhanhui: If you don't mind, I can add a commit to do this.

slp: I'm aware of that pattern, it's used in almost every device in libkrun. It's also completely unnecessary here.

teohhanhui:
I'll need more time to review the rest of the code, if that's alright.

crates/krun-server/src/bin/krun-server.rs
teohhanhui: > It's also completely unnecessary here.

I think abruptly stopping the server thread could still cause us issues. For example, the `"OK"` response not being sent despite the command being spawned successfully. I'd have to trace along for any other unintended consequences...

https://github.com/rust-lang/miri/issues/1371#issuecomment-754834840

crates/krun-server/src/cli_options.rs
```diff
@@ -0,0 +1,38 @@
+use bpaf::{any, construct, env, positional, OptionParser, Parser};
+
+#[derive(Clone, Debug)]
+pub struct Options {
+    pub server_port: u32,
```
teohhanhui, line 5:
```suggestion
    pub port: u32,
```

crates/krun-server/src/cli_options.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub command: String,
+    pub command_args: Vec<String>,
+}
+
+pub fn options() -> OptionParser<Options> {
+    let server_port = env("KRUN_SERVER_PORT")
+        .help("TCP port to listen for command launch requests")
+        .argument::<u32>("KRUN_SERVER_PORT")
+        .fallback(3333);
```
teohhanhui, line 14:
```suggestion
    let port = long("port")
        .short('p')
        .env("KRUN_SERVER_PORT")
        .help("TCP port to listen for command launch requests")
        .argument("PORT")
        .fallback(3333)
        .display_fallback();
```

crates/krun-server/src/cli_options.rs
```diff
@@ -0,0 +1,38 @@
...
+    let command = positional("COMMAND");
+    let command_args = any::<String, _, _>("COMMAND_ARGS", |arg| {
+        (!["--help", "-h"].contains(&&*arg)).then_some(arg)
+    })
+    .many();
+
+    construct!(Options {
+        server_port,
```
teohhanhui, line 22:
```suggestion
        port,
```

crates/krun-server/src/cli_options.rs
```diff
@@ -0,0 +1,38 @@
+use bpaf::{any, construct, env, positional, OptionParser, Parser};
```
teohhanhui, line 1:
```suggestion
use bpaf::{any, construct, long, positional, OptionParser, Parser};
```

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,27 @@
...
+        if let Err(err) = start_server(options.server_port) {
```
teohhanhui, line 14:
```suggestion
        if let Err(err) = start_server(options.port) {
```

teohhanhui (commented):
More uniform CLI options...

(If the user decides to introduce a mismatch between `krun --server-port` and `krun-server --port` then tough luck...)

crates/krun-server/src/cli_options.rs
teohhanhui: Uhh... Sorry, I see now that it's quite pointless to have long / short options here.

crates/krun-server/src/cli_options.rs
slp: I'm not entirely convinced about this suggestion. It's likely we'll need to eventually add support for redirecting ports with passt, and I feel the name `server_port` will avoid confusion in the future...

crates/krun-server/src/bin/krun-server.rs
slp: I've just implemented it. To avoid adding new dependencies or complicating the code too much, we're keeping a copy of the RawFd and shutting it down when tearing down things, which causes TcpListener::incoming() to exit, allowing us to wait for and join the server thread successfully.

crates/krun-server/src/cli_options.rs
teohhanhui: Noted.

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -0,0 +1,33 @@
...
+    let listener_fd = listener.as_raw_fd();
+
+    let server_thread = start_server(listener);
+
+    let err = Command::new(&options.command)
+        .args(options.command_args)
+        .status();
+    if let Err(err) = err {
+        println!("Error executing command {}: {}", options.command, err);
+    }
```
teohhanhui, line 25:
```suggestion
    let status = Command::new(&options.command)
        .args(options.command_args)
        .status();
    if let Err(err) = status {
        println!("Error executing command {}: {}", options.command, err);
    }
```

teohhanhui (approved):
Didn't manage to go through all of the code (it's quite a lot), but let's ship it!

slp:
> Didn't manage to go through all of the code (it's quite a lot), but let's ship it!

Yeah, sorry about that. This turned to be a huge PR basically because we're changing a significant part of the original behavior and UX. I hope we don't need to deal with such a large change anymore. Everything else I have line up (including a juicy one that drastically decreases start-up time ;-) is much smaller.

I've just applied your last suggestions and will be merging this one. Thanks a lot for the review!

crates/krun-guest/src/cli_options.rs
teohhanhui: #28

crates/krun-server/src/bin/krun-server.rs
teohhanhui: #28

crates/krun/src/bin/krun.rs
```diff
@@ -317,6 +312,8 @@ fn main() -> Result<()> {
         env.push(s);
     }
 
+    env.push(CString::new(format!("KRUN_SERVER_PORT={}", options.server_port)).unwrap());
```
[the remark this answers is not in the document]
teohhanhui: #28

PR #25, opened 2024-06-02
Redirect socat

slp:
Redirect socat output to avoid breaking console apps.

### DIFF
--- a/crates/krun-guest/src/pulse.rs
+++ b/crates/krun-guest/src/pulse.rs
@@ -1,9 +1,12 @@
+use std::collections::HashMap;
+use std::env;
 use std::fs;
 use std::path::Path;
-use std::process::Command;
+use std::process::{Command, Stdio};
 
 use anyhow::{Context, Result};
 use utils::env::find_in_path;
+use utils::stdio::make_stdout_stderr;
 
 pub fn setup_pulse_proxy<P>(run_path: P) -> Result<()>
 where
@@ -13,6 +16,9 @@ where
     let Some(socat_path) = socat_path else {
         return Ok(());
     };
+
+    let envs: HashMap<String, String> = env::vars().collect();
+    let (stdout, stderr) = make_stdout_stderr(&socat_path, &envs)?;
 
     let run_path = run_path.as_ref();
     let pulse_path = run_path.join("pulse");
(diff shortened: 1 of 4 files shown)


teohhanhui:
Is the inclusion of this commit intentional?

> Introduce a client-server model

teohhanhui:
`git checkout origin/main -- Cargo.lock`?

crates/krun-guest/src/pulse.rs
```diff
@@ -1,14 +1,17 @@
...
+use utils::{env::find_in_path, stdout::create_stdout_stderr};
 
 pub fn setup_pulse_proxy(run_path: PathBuf) -> Result<()> {
     if let Some(socat_path) = find_in_path("socat")? {
         let pulse_path = run_path.join("pulse");
         fs::create_dir(&pulse_path)
             .context("Failed to create `pulse` directory in `XDG_RUNTIME_DIR`")?;
         let pulse_path = pulse_path.join("native");
+
+        let (stdout, stderr) = create_stdout_stderr("socat")?;
```
teohhanhui, line 13: WDYT about just using [`piped`](https://doc.rust-lang.org/std/process/struct.Stdio.html#method.piped)? Or is this important for debuggability?

slp: I think it could be useful in case something goes wrong with the pulseaudio bridge.

slp:
> `git checkout origin/main -- Cargo.lock`?

Ouch.

crates/utils/src/lib.rs
```diff
@@ -1,2 +1,3 @@
 pub mod env;
 pub mod launch;
+pub mod stdout;
```
teohhanhui, line 3:
```suggestion
pub mod stdio;
```

teohhanhui:
> `git checkout origin/main -- Cargo.lock`?

Oops... Actually, the `Cargo.lock` in `main` is incorrect. That's why every time we do `cargo clippy` or `cargo build` it'll change `Cargo.lock`...

slp:
Addressed suggestions and rebased on the latest HEAD.

teohhanhui (approved):
LGTM

PR #26, opened 2024-06-02
Drop support for TSI networking

slp:
The use case targeted by krun is not compatible with TSI networking, so  drop it as an option to avoid unnecessary confusion.


teohhanhui (approved):
LGTM

PR #28, opened 2024-06-05
Clean up some nits and fix some minor bugs

teohhanhui:


crates/krun-guest/src/bin/krun-guest.rs
```diff
@@ -14,16 +18,16 @@ fn main() -> Result<()> {
     let options = options().run();
 
     {
-        // Raise RLIMIT_NOFILE to the maximum allowed. This is required for wine's esync to work.
-        // See https://github.com/lutris/docs/blob/master/HowToEsync.md
+        // Raise RLIMIT_NOFILE to the maximum allowed. This is required for wine's esync
+        // to work. See https://github.com/lutris/docs/blob/master/HowToEsync.md
```
teohhanhui, line 22: Urgh...

crates/krun-guest/src/fex.rs
```diff
@@ -1,10 +1,17 @@
...
-const FEX_X86_64_BINFMT_MISC_RULE: &str = ":FEX-x86_64:M:0:\\x7fELF\\x02\\x01\\x01\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x02\\x00\\x3e\\x00:\\xff\\xff\\xff\\xff\\xff\\xfe\\xfe\\x00\\x00\\x00\\x00\\xff\\xff\\xff\\xff\\xff\\xfe\\xff\\xff\\xff:${FEX_INTERPRETER}:POCF";
+const FEX_X86_BINFMT_MISC_RULE: &str = ":FEX-x86:M:0:\\x7fELF\\x01\\x01\\x01\\x00\\x00\\x00\\x00\\\
+                                        x00\\x00\\x00\\x00\\x00\\x02\\x00\\x03\\x00:\\xff\\xff\\\
+                                        xff\\xff\\xff\\xfe\\xfe\\x00\\x00\\x00\\x00\\xff\\xff\\\
+                                        xff\\xff\\xff\\xfe\\xff\\xff\\xff:${FEX_INTERPRETER}:POCF";
+const FEX_X86_64_BINFMT_MISC_RULE: &str =
+    ":FEX-x86_64:M:0:\\x7fELF\\x02\\x01\\x01\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x02\\\
+     x00\\x3e\\x00:\\xff\\xff\\xff\\xff\\xff\\xfe\\xfe\\x00\\x00\\x00\\x00\\xff\\xff\\xff\\xff\\\
+     xff\\xfe\\xff\\xff\\xff:${FEX_INTERPRETER}:POCF";
```
teohhanhui, line 14: Need to check that this reformatting is okay...

teohhanhui: Confirmed that it's okay.

crates/krun/src/lock.rs
```diff
@@ -110,8 +108,9 @@ fn request_launch(
         }
```
teohhanhui, line 108: This is wrong. Fixing...

teohhanhui: Fixed.

crates/krun-server/src/server.rs
```diff
@@ -52,7 +50,7 @@ fn handle_connection(mut stream: TcpStream) -> Result<()> {
     let mut envs: HashMap<String, String> = env::vars().collect();
 
     let launch = read_request(&stream)?;
-    envs.extend(launch.envs);
+    envs.extend(launch.env);
```
teohhanhui, line 53: Is there a reason for doing this? Doesn't this mean the user cannot have precise control over the env vars for a specific invocation of `krun`?

crates/krun/src/bin/krun.rs
```diff
@@ -208,8 +203,26 @@ fn main() -> Result<()> {
...
+            format!("Failed to process {PROGRAM:?} path as it contains invalid UTF-8")
+        })?)
+        .with_context(|| {
+            format!("Failed to process {PROGRAM:?} path as it contains NUL character")
+        })?
+    };
+    let krun_server_path = {
+        const PROGRAM: &str = "krun-server";
+        let path = find_krun_exec(PROGRAM)?;
+        CString::new(path.to_str().with_context(|| {
```
slp, line 219: Why can't we do the `CString` conversion in `find_krun_exec`.

teohhanhui: Sure we could, but we don't always want a `CString`. Returning a `PathBuf` for a path is less surprising / more flexible?

crates/krun-server/src/server.rs
slp: Running graphical apps requires a number of `envs` exported by `sommelier`. The idea is that users are able to export _additional_ environment variables, as otherwise the UX would be pretty bad. If there's a demand for that, we could add another option in the future to just export those variables.

teohhanhui: Yeah, I kinda realized this after a while... Haha.

crates/krun/src/bin/krun.rs
slp: The two uses we have right now need a `CString`. I think it's fine to keep it as it, and generalize it if/when we need to use it in more places.

teohhanhui: Alright.

teohhanhui:
Done. And removed unnecessary cloning.

slp:
> Done. And removed unnecessary cloning. (`lock_or_connect` now takes and gives back ownership in the case of `LockAcquired`)
> 
> Hmm... Actually `lock_or_connect` does not adequately describe what the function actually does... Should it be `launch_or_lock`?
> 
> **EDIT**: Renamed

Could you please rename it to `lock_or_launch`? I think it reflects better that (except it `KRUN_SERVER_PORT` env is present) it'll first try to lock and only launch if it can't acquire the lock.

slp:
LGTM. I also tested this and didn't find any issues. I've just suggested a very minor change, but otherwise I'm fine merging it.

teohhanhui:
> Could you please rename it to `lock_or_launch`? I think it reflects better that (except it `KRUN_SERVER_PORT` env is present) it'll first try to lock and only launch if it can't acquire the lock.

I get what you mean, but... This seems more natural to me: we first attempt to launch using an existing `krun` instance, and only continuing if the launch did not happen. That also fits with the usual "right" (`Err`) case of giving back ownership. Except we're not using `Err` here for `anyhow` reasons...

PR #31, opened 2024-06-06
Various fixes for launching processes on a running instance.

slp:


crates/krun-server/src/server.rs
```diff
@@ -46,6 +46,27 @@ fn read_request(mut stream: &TcpStream) -> Result<Launch> {
...
+            let path_stderr = base.join(format!("krun-{filename}-{ts}.stderr"));
+            return Ok((
+                File::create_new(path_stdout)?.into(),
+                File::create_new(path_stderr)?.into(),
+            ));
+        }
+    }
+
+    Ok((Stdio::null(), Stdio::null()))
+}
```
teohhanhui, line 68:
```suggestion
fn make_stdout_stderr<P>(command: P, envs: &HashMap<String, String>) -> Result<(Stdio, Stdio)>
where
    P: AsRef<Path>,
{
    let command = command.as_ref();
    let Some(filename) = command.file_name() else {
        return Ok((Stdio::null(), Stdio::null()));
    };
    let filename = filename
        .to_str()
        .context("Failed to process command as it contains invalid UTF-8")?;
    let base = if envs.contains_key("XDG_RUNTIME_DIR") {
        Path::new(&envs["XDG_RUNTIME_DIR"])
    } else {
        Path::new("/tmp")
    };
    let ts = SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis();
    let path_stdout = base.join(format!("krun-{filename}-{ts}.stdout"));
    let path_stderr = base.join(format!("krun-{filename}-{ts}.stderr"));
    Ok((
        File::create_new(path_stdout)?.into(),
        File::create_new(path_stderr)?.into(),
    ))
}
```

crates/krun/src/launch.rs
```diff
@@ -22,6 +23,29 @@ pub enum LaunchResult {
     },
 }
 
+#[derive(Debug)]
+enum LaunchError {
```
teohhanhui, line 27: We should `impl Error` too?

crates/krun-server/src/server.rs
slp: I'm fine with using a bound trait, but I think if the string conversion fails we should also return `(Stdio::null(), Stdio::null())`. I mean, is not much worse that `Path::file_name()` failing, and we're doing that for it.

teohhanhui: On second thought, could we not just error out if `file_name` is `None`? In which cases are we expecting that?

crates/krun/src/launch.rs
```diff
@@ -112,26 +147,37 @@ fn request_launch(
     command: PathBuf,
     command_args: Vec<String>,
     env: HashMap<String, String>,
-) -> Result<()> {
-    let mut stream = TcpStream::connect(format!("127.0.0.1:{server_port}"))?;
+) -> std::result::Result<(), LaunchError> {
```
teohhanhui, line 150: If `LaunchError` impls `Error` we could just use `anyhow::Result` here as usual...

crates/krun-server/src/server.rs
```diff
@@ -56,24 +77,12 @@ fn handle_connection(mut stream: TcpStream) -> Result<()> {
...
-        };
-        let ts = SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis();
-        let path_stdout = base.join(format!("krun-{}-{ts}.stdout", command.display()));
-        let path_stderr = base.join(format!("krun-{}-{ts}.stderr", command.display()));
-        (
-            File::create_new(path_stdout)?,
-            File::create_new(path_stderr)?,
-        )
-    };
+    let (stdout, stderr): (Stdio, Stdio) = get_stdout_stderr(&command, &envs)?;
```
teohhanhui, line 80: Unnecessary type annotation?

crates/krun/src/launch.rs
```diff
@@ -22,6 +23,29 @@ pub enum LaunchResult {
...
+        match *self {
+            Self::Connection(ref err) => {
+                write!(f, "Can't connect to krun server: {err}")
+            },
+            Self::Json(ref err) => {
+                write!(f, "Can't transform arguments into JSON: {err}")
+            },
+            Self::Server(ref err) => {
+                write!(f, "krun server returned an error: {err}")
+            },
```
teohhanhui, line 44:
```suggestion
            Self::Connection(ref err) => {
                write!(f, "could not connect to krun server: {err}")
            },
            Self::Json(ref err) => {
                write!(f, "could not serialize into JSON: {err}")
            },
            Self::Server(ref err) => {
                write!(f, "krun server returned an error: {err}")
            },
```

> Error messages are typically concise lowercase sentences without trailing punctuation

https://doc.rust-lang.org/std/error/trait.Error.html

teohhanhui (approved):
LGTM once the above points are addressed.

crates/krun-server/src/server.rs
teohhanhui: > Returns `None` if the path terminates in `...`

https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name

I think we can just error.

slp: Applied the change, bailing out if file_name is None.

crates/krun/src/launch.rs
slp: Done.

crates/krun/src/launch.rs
slp: Done.

crates/krun-server/src/server.rs
slp: Removed

slp:
@teohhanhui PTAL

crates/krun/src/launch.rs
```diff
@@ -22,6 +24,30 @@ pub enum LaunchResult {
     },
 }
 
+#[derive(Debug)]
+enum LaunchError {
+    Connection(std::io::Error),
+    Json(serde_json::Error),
+    Server(String),
+}
+impl Error for LaunchError {}
```
teohhanhui, line 33: Missing blank line before.

teohhanhui (approved):
LGTM

slp:
Thanks!

crates/krun/src/launch.rs
```diff
@@ -22,6 +24,31 @@ pub enum LaunchResult {
...
 }
 
+#[derive(Debug)]
+enum LaunchError {
+    Connection(std::io::Error),
+    Json(serde_json::Error),
+    Server(String),
+}
+
+impl Error for LaunchError {}
```
teohhanhui, line 34: Note to self:

> If there are no items, format the impl (including its `{}`) on a single line.

https://doc.rust-lang.org/nightly/style-guide/items.html#impls

The currently commented out rule we're using for `cargo +nightly fmt` is wrong:

https://github.com/slp/krun/blob/d0179f3c4ab0eaa07a8813a772db805bf5be9965/rustfmt.toml#L3

PR #32, opened 2024-06-07
Raise RAM limits

slp:


teohhanhui:
Does this require `libkrunfw` changes?

slp:
> This would require `libkrun` / `libkrunfw` changes, right?
> 
> [#11 (comment)](https://github.com/slp/krun/pull/11#discussion_r1617256974)

No, I forgot I raised the RAM limit years ago. The vCPU limit is there though, and can't be raised until we implement GICv3 in libkrun.

teohhanhui:
> The vCPU limit is there though

As in it's limited to 8 vCPUs? We should check that in our code then... I can do a follow-up PR if that's the case.

crates/krun/src/bin/krun.rs
```diff
@@ -103,7 +103,7 @@ fn main() -> Result<()> {
         }
         debug!(cpuset:?; "sched_setaffinity");
         sched_setaffinity(None, &cpuset).context("Failed to set CPU affinity")?;
-        // Configure the number of vCPUs and the amount of RAM (max 16384 MiB).
+        // Configure the number of vCPUs and the amount of RAM (max 32768 MiB).
```
teohhanhui, line 106: This comment here is wrong if we remove the `guard` in `cli_options`. The user could simply specify more than 32768 MiB.

alyssarosenzweig (approved):

teohhanhui:
Possible fallout: https://gist.github.com/teohhanhui/042a395010d9946ceee14768736e3780?permalink_comment_id=5197857#gistcomment-5197857

Haven't looked into it yet, but perhaps we're hitting a limit in libkrun? It could be something that's already fixed upstream though...

PR #34, opened 2024-06-09
Make krun-server wait for child processes

### LINKED ISSUE #33: krun-server should outlive first command
(Continuing from https://github.com/slp/krun/pull/23#discussion_r1624008269)

With some real-world testing, it's entirely jarring behaviour from a user perspective.

Steps for reproduction:
1. Launch `krun FEXBash`
2. Launch `krun FEXInterpreter steam`
3. `exit` from `FEXBash`

Expected result:
Steam should continue running.

Actual / observed behaviour:
Steam is killed abruptly.

teohhanhui:
Continue to accept and handle incoming connections, until the server is idle and all child processes have exited.

Fixes #33

Tested working with all of the scenarios mentioned in the linked issue, and confirmed that the server exits correctly once all the child processes have exited (doesn't seem to run into deadlocks).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.22.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -47,7 +62,7 @@ version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a64c907d4e79225ac72e2a354c9ce84d50ebb4586dee56c82b3ee73004f537f5"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -57,7 +72,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "61a38449feb7068f52bb06c12759005cf459ee52bb4adc1d5a7c4322d716fb19"
 dependencies = [
  "anstyle",
- "windows-sys",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -67,10 +82,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25bdb32cbbdce2b519a9cd7df3a678443100e265d5e25ca763b7572a5104f5f3"
 
 [[package]]
-name = "autocfg"
-version = "1.3.0"
+name = "backtrace"
+version = "0.3.72"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
+checksum = "17c6a35df3749d2e8bb1b7b21a976d82b15548788d2735b9d82f329268f71a11"
+dependencies = [
+ "addr2line",
+ "cc",
+ "cfg-if",
+ "libc",
+ "miniz_oxide",
+ "object",
+ "rustc-demangle",
+]
 
 [[package]]
 name = "bindgen"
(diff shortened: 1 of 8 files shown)


Cargo.lock
```diff
@@ -66,12 +66,6 @@ version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25bdb32cbbdce2b519a9cd7df3a678443100e265d5e25ca763b7572a5104f5f3"
 
-[[package]]
```
teohhanhui, line 69: I have no idea what's going on here anymore...

teohhanhui: Oh, it's because we're no longer using the `socket` feature of `nix` crate...

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -1,32 +1,83 @@
...
+                },
+                Err(TryRecvError::Disconnected) => {
+                    panic!("child_tx in server thread should not be gone");
+                },
+            }
+        };
+        let child_pid = child.id();
+        let status = child
+            .wait()
+            .context("Failed to wait for child process to exit")?;
```
teohhanhui, line 68: Hmm... Maybe we shouldn't propagate this error. It's more important to wait for the other child processes, rather than dying here...

teohhanhui: Done. Made the failure to run main command not error as well, since we might already have child processes launched / being launched at that point.

crates/krun-server/src/server.rs
```diff
@@ -43,32 +68,34 @@ fn read_request(mut stream: &TcpStream) -> Result<Launch> {
...
         .stdin(Stdio::null())
         .stdout(stdout)
         .stderr(stderr)
-        .spawn();
-    if let Err(err) = err {
-        let msg = format!("Failed to execute command {command:?}: {err}");
+        .spawn()
+        .with_context(|| format!("Failed to execute {command:?} as child process"));
+    if let Err(err) = &spawn_result {
+        let msg = format!("{err:?}");
```
teohhanhui, line 93: It is okay for us to spam the client with the whole backtrace, right? :rofl:

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -1,32 +1,96 @@
...
+            eprintln!(
+                "Failed to execute {:?} as child process: {err}",
+                options.command
+            );
+        },
     }
 
-    Ok(())
+    println!("Waiting for other commands launched through this krun server to exit...");
+    println!("Press Ctrl+C to force quit");
```
slp, line 54: Ideally, this should only be printed if we're really going to wait on some child.

slp:
Frankly, I'm not a big fan of this approach. Here we're sacrificing joining cleaning the thread and introducing an overly complex pattern for what we really need to do. And, what's worse, we're not reaping the children until the process is going to end, meaning we're leaving zombie processes in the system.

The hardest part to solve is the latter. We basically have 3 options:

1. A signal handler for SIGCHLD with a global Arc<Mutex<Vec<Child>>> that's kept updated by everyone. When exiting, wait on whatever children are still on the Vec. Not a big fan of this one either.
2. Spawn a thread with each children and have it wait on the child. Put all threads in a Arc<Mutex<Vec<JoinHandle>>>. On exit, join both the the server thread and all process monitoring threads.
3. Bite the bullet and use Tokyo, which allows dealing with children (and the server) in an asynchronous fashion.

I think I would favor 2, but I suspect we're going to need Tokyo sooner than later (a crate I'm looking for changing the way in which the network interface is configured requires it, but maybe there's some other option), so I'm fine with either.

teohhanhui:
@slp I'd favour tokio. But I think the implementation will only end up being more complex, not less. (I think it's pretty much unavoidable.)

Would you be okay with merging this as-is? I think it really helps with usability despite the issues you've identified above.

crates/krun-server/src/bin/krun-server.rs
teohhanhui: We could print the message the first time we're going to `wait` for a child. WDYT?

teohhanhui: Done.

slp:
> @slp I'd favour tokio. But I think the implementation will only end up being more complex, not less. (I think it's pretty much unavoidable.)

It'll be more complex internally, but the code for us would be simpler thanks to tokio/async's sugar syntax.

> Would you be okay with merging this as-is? I think it really helps with usability despite the issues you've identified above. I'll work on a follow-up PR.

Since we're not in a hurry to provide this to users, I think we should take our time to do it right. I might be able to work on a first implementation of tokio later this week, if you don't beat me to it ;-)

teohhanhui:
> Since we're not in a hurry to provide this to users, I think we should take our time to do it right. I might be able to work on a first implementation of tokio later this week, if you don't beat me to it ;-)

Alright. I'll start working on this tomorrow. :smile:

teohhanhui:
Done. :rocket:

crates/krun-server/src/bin/krun-server.rs
```diff
@@ -1,32 +1,94 @@
...
-    }
+    let mut server_died = false;
+    let mut command_exited = false;
 
-    Ok(())
+    loop {
+        tokio::select! {
+            res = &mut server_handle, if !server_died => {
+                // If an error is received here, accepting connections from the
+                // TCP listener failed due to non-transient errors and the
```
teohhanhui, line 40: TODO: This is a lie. We're logging and ignoring all connection errors at this time... We should continue on transient errors but bail out on non-transient errors.

https://book.async.rs/patterns/accept-loop

crates/krun-server/src/server.rs
```diff
@@ -43,32 +135,34 @@ fn read_request(mut stream: &TcpStream) -> Result<Launch> {
...
         .stderr(stderr)
-        .spawn();
-    if let Err(err) = err {
-        let msg = format!("Failed to execute command {command:?}: {err}");
-        stream.write_all(msg.as_bytes()).ok();
+        .spawn()
+        .with_context(|| format!("Failed to execute {command:?} as child process"));
+    if let Err(err) = &res {
+        let msg = format!("{err:?}");
+        stream.write_all(msg.as_bytes()).await.ok();
```
teohhanhui, line 161: `write_all` is not cancellation safe. Expect Heisenbugs!

slp:
Woah, that was fast! I think we could refactor `start_server`'s loop in various methods to make it look better, but I don't want to delay this PR just for that. Let's get it merged!

slp (approved):

teohhanhui:
Working on the refactoring of `start_server` :D

PR #35, opened 2024-06-13
Refactor start_server function

teohhanhui:


crates/krun-server/src/server.rs
```diff
@@ -1,124 +1,123 @@
...
+        self.state_tx.send_if_modified(|state| {
+            if state.connection_idle == connection_idle {
+                return false;
+            }
+            state.connection_idle = connection_idle;
+            true
+        });
+    }
+
+    fn set_child_processes(&self, child_processes: usize) {
```
teohhanhui, line 113: It is of course unnecessary to take `child_processes` as a parameter. But it reads better?

slp (approved):

PR #38, opened 2024-06-20
Passthru NixOS system & OpenGL Driver

### LINKED ISSUE #37: Issues with `/run` on NixOS
On NixOS, we use `/run` for many runtime things such as symlinking the current system state to `/run/current-system`. This means the system "installed" packages will not be accessible unless `/run/current-system` is available. We also have the OpenGL drivers symlinked to `/run/opengl-driver`. Without that symlink being available, GPU related access via Mesa is not possible. Since `/run` isn't shared to `krun`, this means system packages and OpenGL are both not available.

Possible solution is to check if `NIXOS_SYSTEM` is an env var set and to mount the value of it to `/run/current-system`. A similar solution could be done for OpenGL.

RossComputerGuy:
Fixes #37 by passing thru the NixOS system and OpenGL driver.

### DIFF
--- a/crates/krun-guest/src/mount.rs
+++ b/crates/krun-guest/src/mount.rs
@@ -1,9 +1,10 @@
-use std::fs::File;
+use std::fs::{File, create_dir_all};
+use std::env::var_os;
 use std::os::fd::AsFd;
 
 use anyhow::{Context, Result};
 use rustix::fs::CWD;
-use rustix::mount::{mount2, move_mount, open_tree, MountFlags, MoveMountFlags, OpenTreeFlags};
+use rustix::mount::{mount2, move_mount, mount_recursive_bind, open_tree, MountFlags, MoveMountFlags, OpenTreeFlags};
 
 pub fn mount_filesystems() -> Result<()> {
     mount2(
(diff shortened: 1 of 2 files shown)


RossComputerGuy:
Gets as far as:
```
sommelier_scope_timer: init wayland channel: 0.000250 seconds
DRM_IOCTL_ASAHI_GET_PARAMS failed: Invalid argument
sommelier_scope_timer: drm device: 0.070801 seconds
sommelier_scope_timer: connect display: 0.000026 seconds
sommelier_scope_timer: client create: 0.000015 seconds
sommelier_scope_timer: display implementation: 0.000004 seconds
sommelier_scope_timer: spawn xwayland: 0.000156 seconds
_XSERVTransSocketUNIXCreateListener: ...SocketCreateListener() failed
_XSERVTransMakeAllCOTSServerListeners: server already running
```
Then hangs, it's progress

slp:
@teohhanhui PTAL.

teohhanhui:
This is the wrong approach.

Read here for the solution:
https://nixos.wiki/wiki/OpenGL

RossComputerGuy:
> This seems like it's maybe the wrong approach?

Yeah this is the first time I coded in Rust and I'm not entirely sure how krun works so this was the approach I could come up with.

> https://nixos.wiki/wiki/OpenGL
> 
> Disclaimer: I have never used NixOS or even Nix. I'm just basing my judgement on the documentation I can find.

New wiki: https://wiki.nixos.org/wiki/OpenGL

`/run/opengl-driver` is a symlink to a directory in `/nix/store` and so is `/run/current-system`. Probably the best way is to read the OpenGL driver path on host and the find if it's a symlink, resolve the symlink of it is, and then pass it through. I'm not sure how `/run/current-system` could be resolved but as it is in `$PATH`, could parse that and then resolve and pass through whatever is a symlink.

asahilina:
I'm torn over whether we should share /run as a whole or not. There are things like /run/media that make sense to share. On the other hand, that doesn't work even if you *do* mount it right now due to missing POSIX ACL support...

One thing we do have since a recent change is `/run/muvm-host`. That exposes the entire host FS without any overlaid mounts. A simple way of implementing this now would be to check if `/run/muvm-host/{opengl-driver,current-system}` exist and if they do, symlink them back into `/run`. This could be implemented with a list of things in `/run` to symlink so it can be easily extended in the future, such as also doing `/run/media` once that otherwise works.

RossComputerGuy:
> One thing we do have since a recent change is `/run/muvm-host`. That exposes the entire host FS without any overlaid mounts. A simple way of implementing this now would be to check if `/run/muvm-host/run/{opengl-driver,current-system}` exist and if they do, symlink them back into `/run`.

Yeah, I think that might be a good way.

> Aside: I think it's worth implementing a feature to run a user-specified script as root on VM startup. There are going to be people with custom setups who want to do some mount management or other root-requiring tasks on VM startup.

True, some sort of pre and post run scripts that could be specific or pulled from home might be good as well.

nrabulinski:
This PR is obsolete now

PR #41, opened 2024-06-28
Add support for input device passthrough

WhatAmISupposedToPutHere:
Use [hidpipe](https://github.com/WhatAmISupposedToPutHere/hidpipe) to pass through input devices such as gamepads.

### DIFF
--- a/crates/krun-guest/src/bin/krun-guest.rs
+++ b/crates/krun-guest/src/bin/krun-guest.rs
@@ -12,6 +12,7 @@ use krun_guest::sommelier::exec_sommelier;
 use krun_guest::user::setup_user;
 use log::debug;
 use rustix::process::{getrlimit, setrlimit, Resource};
+use utils::env::find_in_path;
 
 fn main() -> Result<()> {
     env_logger::init();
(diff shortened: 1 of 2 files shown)


crates/krun-guest/src/bin/krun-guest.rs
```diff
@@ -48,6 +49,11 @@ fn main() -> Result<()> {
 
     setup_pulse_proxy(run_path)?;
 
+    Command::new("/usr/lib/systemd/systemd-udevd").spawn()?;
```
teohhanhui, line 52: @slp Do we want to go this route?

slp: @teohhanhui it's either that or having to implement something equivalent ourselves, which seems an overkill to me. Do you foresee any potential side effects?

teohhanhui: No, if anything, I'm happy to have `systemd-udevd` 😄

But we should spawn this much earlier so that `dhcpcd` can make use of it, for example. (If that's expected to work...)

WhatAmISupposedToPutHere: Should i move this call earlier then?

teohhanhui: Sure.

slp (approved):
LGTM, let's get it merged.

PR #44, opened 2024-07-06
Support native X11 forwarding

asahilina:
If the host has DISPLAY set, parse it and forward that UNIX socket to the guest via vsock, then forward :1 on the guest to this vsock, and finally set HOST_DISPLAY=:1 on the guest.
    
Apps that want to directly talk to X11 on the host (instead of going through sommelier) should set DISPLAY=$HOST_DISPLAY and LIBGL_ALWAYS_SOFTWARE=true. This setup allows X11 apps in the host and the guest to cooperate (e.g. XEmbed) but requires software rendering inside the guest.
    
TODO: Figure out why things hang without LIGL_ALWAYS_SOFTWARE=true. Ideally DRI3 should fail and fall back to software automatically.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,6 +130,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3280efcf6d66bc77c2cf9b67dc8acee47a217d9be67dd590b3230dffe663724d"
 
 [[package]]
+name = "byteorder"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
+
+[[package]]
 name = "bytes"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -301,6 +307,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "bpaf",
+ "byteorder",
  "env_logger",
  "log",
  "nix",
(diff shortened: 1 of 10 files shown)


slp:
Code LGTM. Tested on Fedora Asahi, works well.

@asahilina do we have a specific use case for this?

@teohhanhui PTAL

asahilina:
This is for running VST (or whatever) audio plugins in krun embedded into a DAW outside of krun, but I haven't gotten to integrating support into yabridge yet!

The more general use case is "proprietary app doing X11 shenanigans which doesn't really need GPU acceleration" which probably covers quite a few niche use cases. Eventually once the client/server stuff is a bit more fleshed out, I figure we should have some kind of --x11-passthrough flag in krun directly so users can request the right environment variables to be set for specific apps, as needed.

One example more familiar to people is running stuff like Discord or Zoom in krun under emulation, and sharing your screen. I haven't tested this at all, but I *think* it should work with X11 passthrough thanks to the [xwayland video bridge](https://planet.kde.org/david-edmundson-2023-03-22-fixing-wayland-xwayland-screen-casting/).

As a more short-term use case, the sommelier stuff is still kind of broken as far as window management (e.g. missing decorations, popup menu problems, refresh issues) for anything other than fullscreen games, so even when direct X11 interaction is not necessary, this is going to be a better experience running many non-GPU-intensive X11 desktop apps today. I assume this will improve over time, but having this option right now makes krun immediately more realistic to use on Asahi with the whole FEX stuff for running x86 X11 apps that aren't games.

(It would be very nice to get GPU acceleration working with this use case of course, but I have no idea how big a can of worms that is...)

slp (commented):
LGTM.

slp (approved):

PR #45, opened 2024-07-06
env: Use generic apple,arm-platform compatible to check for Apple SoCs

asahilina:
This avoids having to explicitly list all current and future devices.

### DIFF
--- a/crates/krun/src/env.rs
+++ b/crates/krun/src/env.rs
@@ -20,20 +20,7 @@ const WELL_KNOWN_ENV_VARS: [&str; 5] = [
 ];
 
 /// See https://github.com/AsahiLinux/docs/wiki/Devices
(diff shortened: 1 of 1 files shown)


teohhanhui (approved):
LGTM

I guess that means it's expected that all Apple Silicon systems will have `asahi` mesa driver support? 🙈

asahilina:
At least right now that's the plan ^^

Of course in the future there will probably be some generation that warrants a new driver, but I think it makes more sense to deal with that when the time comes instead of being too conservative right now. At least up to M4 we definitely expect to share the driver. If we do end up with a split later on, it would probably make more sense to do something with the GPU device compatible or something via the kernel driver, so we can still avoid having to list every SoC explicitly.

PR #47, opened 2024-07-06
Add dynamic UNIX socket forwarding

asahilina:
Unconditionally forward a range of ports to sockets named `$XDG_RUNTIME_DIR/krun/socket/port-$PORT`. This allows applications to dynamically bind to sockets at those paths and have them forwarded to the guest as vsock ports, after the krun VM is already running.

The port range is currently hardcoded as 50000..50200.

### DIFF
--- a/crates/krun/src/bin/krun.rs
+++ b/crates/krun/src/bin/krun.rs
@@ -179,6 +179,25 @@ fn main() -> Result<()> {
                 return Err(err).context("Failed to configure vsock for pulse socket");
             }
         }
+
+        let socket_dir = Path::new(&run_path).join("krun/socket");
+        std::fs::create_dir_all(&socket_dir)?;
+        // Dynamic ports: Applications may listen on these sockets as neeeded.
+        for port in 50000..50200 {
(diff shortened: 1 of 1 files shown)


teohhanhui:
Would it make more sense for the port range to be configurable? The app will have to do directory traversal to figure out which ports are forwarded anyway, right? Otherwise what should happen when a port is already in use?

asahilina:
I think we need some way to tell apps what port range is available, I was thinking maybe a flag file in the directory? Apps would have to then pick the first nonexistent socket/port to bind to.

slp:
Looks good. We should probably make the port range configurable via args with good defaults. As for how to coordinate port numbers across guest and host, giving that we expect most application launches to be initiated from the host, a reasonable approach could be delegating this responsibility to the application itself, like this:

- The user launches the app.
- The app looks for an non-existing socket file in the well-known directory.
- The app launches the host-side executable which creates the unix socket in the selected path.
- The app launches the guest-side executable using `krun` and passing an environment variable indicating the vsock port to connect to.

If this approach works for us, the only change in this PR would be making the port range configuration (and even than isn't a strong requirement from me).

PR #48, opened 2024-07-06
Add interactive launch mode.

WhatAmISupposedToPutHere:
After launching a process via the server, krun immediately closes and the
launched process' std{in,out,err} go nowhere. Add a --interactive flag that
allocates a pseudo-terminal, routes it to the host and connects to it.

Somewhat like '-it' in docker/podman exec.

### DIFF
--- a/crates/muvm/src/bin/muvm.rs
+++ b/crates/muvm/src/bin/muvm.rs
@@ -297,6 +297,7 @@ fn main() -> Result<()> {
                 return Err(err).context("Failed to configure vsock for pulse socket");
             }
         }
+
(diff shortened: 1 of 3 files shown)


teohhanhui:
@WhatAmISupposedToPutHere Judging from the commit history, does this PR depend on #47? That's still in draft status anyway.

WhatAmISupposedToPutHere:
Technically yes, but i am not super sure why that one is marked as draft.

slp:
I think we should get this one merged for the next release. Can you please rebase it?

slp (approved):
Reviewed and tested, thanks!

PR #51, opened 2024-08-17
Use uuidv7 for stdout / stderr files

teohhanhui:
It's too easy to hit the same millisecond...

Example of error message:

```
Error: File exists (os error 17)
```


slp (approved):
LGTM.

PR #53, opened 2024-09-01
Skip using resolvconf (more dhcp/dns fixes)

WhatAmISupposedToPutHere:
On systems with systemd-resolved installed, dhcpcd tries to use resolvconf to configure the dns server. Naturally this fails, as we do not have systemd &co inside the vm.

### DIFF
--- a/crates/krun-guest/src/net.rs
+++ b/crates/krun-guest/src/net.rs
@@ -28,7 +28,7 @@ pub fn configure_network() -> Result<()> {
     };
     if let Some(dhcpcd_path) = dhcpcd_path {
         let output = Command::new(dhcpcd_path)
-            .args(["-M", "--nodev", "eth0"])
+            .args(["-M", "eth0", "-e", "resolvconf=does-not-exist"])
             .output()
             .context("Failed to execute `dhcpcd` as child process")?;
         debug!(output:?; "dhcpcd output");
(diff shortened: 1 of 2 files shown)


crates/krun-guest/src/net.rs
```diff
@@ -28,7 +28,7 @@ pub fn configure_network() -> Result<()> {
     };
     if let Some(dhcpcd_path) = dhcpcd_path {
         let output = Command::new(dhcpcd_path)
-            .args(["-M", "--nodev", "eth0"])
+            .args(["-M", "--nodev", "eth0", "-e", "resolvconf=does-not-exist"])
```
teohhanhui, line 31: By the way, we have `systemd-udevd` now, right? Can you test if it works without `--nodev`?

WhatAmISupposedToPutHere: It does, but needs a few more udev commands. Added those.

n3ph:
In my tests this only works if the DNS resolver I got via DHCP is an public/external one. In case it's my networks gateway I am connected to, I see no DNS requests leaving my hosts interface.

WhatAmISupposedToPutHere:
Yes, that is true, but that is a separate bug, passt directs the requests to the default gateway to the host machine instead which seems to not be running a dns server in your case. (you might be able to fix this by enabling (not just installing) systemd-resolved)

n3ph:
T _think_ `systemd-resolved` is configured and running properly. It seems `passt` forwards the requests to `udp:127.0.0.1:53`, but turns out the stub resolver listens to `127.0.0.53` and `127.0.0.54`. 

I temporarily fixed it by adding an additional listener in `/etc/systemd/resolved.conf`:
```
[Resolve]
DNSStubListenerExtra=udp:127.0.0.1:53 
```

WhatAmISupposedToPutHere:
Superseded by #111

PR #54, opened 2024-09-01
Support being used as a binfmt_misc handler.

WhatAmISupposedToPutHere:
This series allows krun to be used as a binfmt_misc handler: it uses interactive launch mode to make sure stdin/stdout are going where users expect, passes through the current working directory, expands the number of environment variables that are passed through, and daemonizes the vm process, so a process that waits on the first process launched via krun is not prevented from progressing by others.


m1m1k4tz:
It would be nice to have this for box64 it runs the steam app via box32 now. For it to work on muvm there needs to be some work on binfmt and also some GPU fixes.

WhatAmISupposedToPutHere:
> It would be nice to have this for box64 it runs the steam app via box32 now. For it to work on muvm there needs to be some work on binfmt and also some GPU fixes.

I am not sure as to how this is relevant to the current PR, this one is about configuring the host to start muvm when a foregin binary is launched, not about what happens inside. If you wish to use box64 instead, the relevant file is https://github.com/AsahiLinux/muvm/blob/main/crates/muvm/src/guest/fex.rs . But anyway, thank you for reminding me that i probably should respin this patchset, since a certain issue that prevented it from being merged last time is resolved now.

nrabulinski:
@WhatAmISupposedToPutHere do you have the time to work on this? If not, would you be fine with me taking your changes and rebasing them? (Obviously with the author field being credited to you)

WhatAmISupposedToPutHere:
Sure, go ahead

PR #55, opened 2024-09-03
Refactor project to simplify publishing and packaging

slp:
This chanage aims at simplify publishing `krun` in crates.io. Since
    krun, krun-guest and krun-server are tightly coupled, I don't think it's
    worth treating them independently. Refactoring them to be modules within
    krun, while specifying them as additional binaries, simplifies
    publishing and packaging without significant drawbacks.

    We keep krun-sys as an independent crate to be published by itself, since
    it can potentially be consumed by other programs for linking against
    libkrun.


crates/krun/Cargo.toml
```diff
@@ -1,24 +1,43 @@
...
-rustix = { workspace = true, features = ["process", "std", "use-libc-auxv"] }
-serde = { workspace = true, features = [] }
-serde_json = { workspace = true, features = ["std"] }
-utils = { workspace = true, features = [] }
+anyhow = { version = "1.0.82", default-features = false, features = ["std"] }
+bpaf = { version = "0.9.11", default-features = false }
+byteorder = { version = "1.5.0", default-features = false, features = ["std"] }
+env_logger = { version = "0.11.3", default-features = false, features = ["auto-color", "humantime", "unstable-kv"] }
+#krun-sys = { version = "1.9.1", default-features = false }
+krun-sys = { path = "../krun-sys", default-features = false }
```
teohhanhui, line 17: This kind of use case is explicitly supported: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#local-paths-in-published-crates

```suggestion
krun-sys = { path = "../krun-sys", version = "1.9.1", default-features = false, features = [] }
```

crates/krun/Cargo.toml
```diff
@@ -1,24 +1,43 @@
...
+nix = { version = "0.28.0", default-features = false, features = ["user"] }
+rustix = { version = "0.38.34", default-features = false, features = ["fs", "mount", "process", "std", "system", "use-libc-auxv"] }
+serde = { version = "1.0.203", default-features = false, features = ["derive"] }
+serde_json = { version = "1.0.117", default-features = false, features = ["std"] }
+tempfile = { version = "3.10.1", default-features = false }
+tokio = { version = "1.38.0", default-features = false, features = ["io-util", "macros", "net", "process", "rt-multi-thread", "sync"] }
+tokio-stream = { version = "0.1.15", default-features = false, features = ["net", "sync"] }
+uuid = { version = "1.10.0", default-features = false, features = ["std", "v7"] }
+
+#utils = { path = "crates/utils", default-features = false }
```
teohhanhui, line 28: This can just be removed, right?

crates/krun/Cargo.toml
```diff
@@ -1,24 +1,43 @@
...
-env_logger = { workspace = true, features = ["auto-color", "humantime", "unstable-kv"] }
-krun-sys = { workspace = true, features = [] }
-log = { workspace = true, features = ["kv"] }
-nix = { workspace = true, features = ["user"] }
-rustix = { workspace = true, features = ["process", "std", "use-libc-auxv"] }
-serde = { workspace = true, features = [] }
-serde_json = { workspace = true, features = ["std"] }
-utils = { workspace = true, features = [] }
+anyhow = { version = "1.0.82", default-features = false, features = ["std"] }
+bpaf = { version = "0.9.11", default-features = false }
```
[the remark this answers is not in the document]
slp: Makes sense, done.

crates/krun/Cargo.toml
slp: Yup, fixed, thanks!

crates/krun/Cargo.toml
slp: Nice, let's do this, thanks!

davide125:
You'll also want to symlink `LICENSE` into `crates/krun-sys` to ensure it gets included in the packaged crate (and optionally add a README or symlink the toplevel one too).

slp:
> You'll also want to symlink `LICENSE` into `crates/krun-sys` to ensure it gets included in the packaged crate (and optionally add a README or symlink the toplevel one too).

Good catch, thanks. It's not even the same license as krun, so we should definitely move things around a bit.

alyssarosenzweig (approved):

alyssarosenzweig (approved):

PR #63, opened 2024-09-26
Pass args and envs using a config file

slp:
When using "krun_set_exec", the arguments and environment variables are passed to the guest using the kernel command line. In addition to being limited in length, this is also only compatible with ASCII strings.

libkrun is also capable or reading both arguments and environment variables from a JSON config file, which doesn't suffer from any of those limitations, so switch to using this approach instead.

Depends on https://github.com/containers/libkrun/pull/220

### DIFF
--- a/crates/krun/src/bin/krun.rs
+++ b/crates/krun/src/bin/krun.rs
@@ -1,4 +1,5 @@
 use std::ffi::{c_char, CString};
+use std::io::Write;
 use std::os::fd::{IntoRawFd, OwnedFd};
 use std::path::Path;
 use std::{cmp, env};
@@ -11,7 +12,7 @@ use krun::launch::{launch_or_lock, LaunchResult};
 use krun::net::{connect_to_passt, start_passt};
 use krun::types::MiB;
 use krun_sys::{
-    krun_add_disk, krun_add_vsock_port, krun_create_ctx, krun_set_exec, krun_set_gpu_options,
+    krun_add_disk, krun_add_vsock_port, krun_create_ctx, krun_set_env, krun_set_gpu_options,
     krun_set_log_level, krun_set_passt_fd, krun_set_root, krun_set_vm_config, krun_set_workdir,
     krun_start_enter, VIRGLRENDERER_DRM, VIRGLRENDERER_THREAD_SYNC,
     VIRGLRENDERER_USE_ASYNC_FENCE_CB, VIRGLRENDERER_USE_EGL,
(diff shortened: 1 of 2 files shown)


alyssarosenzweig:
What if an environment var contains a quote? I don't think that gets escaped here and will presumably blow up in libkrun. Should we use an actual json library perhaps?

slp:
> What if an environment var contains a quote? I don't think that gets escaped here and will presumably blow up in libkrun. Should we use an actual json library perhaps?

We already have serde_json, so let's do that.

crates/krun/src/bin/krun.rs
```diff
@@ -294,90 +310,65 @@ fn main() -> Result<()> {
...
-        CString::new(username).expect("username should not contain NUL character"),
-        CString::new(format!("{}", getuid().as_raw()))
-            .expect("uid should not contain NUL character"),
-        CString::new(format!("{}", getgid().as_raw()))
-            .expect("gid should not contain NUL character"),
+    let mut krun_guest_args: Vec<String> = vec![
+        krun_guest_path,
+        username,
+        format!("{}", getuid().as_raw()),
+        format!("{}", getgid().as_raw()),
```
teohhanhui, line 317: I think our MSRV allows this:

```suggestion
        format!("{uid}", uid = getuid().as_raw()),
        format!("{gid}", gid = getgid().as_raw()),
```

crates/krun/src/bin/krun.rs
```diff
@@ -294,90 +310,65 @@ fn main() -> Result<()> {
...
+        krun_config.args.push(arg);
+    }
+    for (key, value) in env {
+        krun_config.envs.push(format!("{}={}", key, value));
+    }
+    let krun_config = KrunBaseConfig {
+        config: krun_config,
     };
 
+    let mut config_file = NamedTempFile::new()
```
teohhanhui, line 348: Add a safety note here about the lifetime requirements of the `config_file` variable? The path will become invalid if this is dropped before krun is done with it.

crates/krun/src/bin/krun.rs
slp: Applied, thanks.

crates/krun/src/bin/krun.rs
slp: Done, thanks.

alyssarosenzweig (approved):
Nice

teohhanhui (approved):

Issue #17, opened 2024-05-30
DNS resolver woes


teohhanhui:
In the VM, my `/etc/resolv.conf` looks like this:
```
; generated by /usr/sbin/dhclient-script
nameserver 192.168.0.1
```
Which is bad, because that's my router's IP address... (And not what's sent as DNS servers via DHCP.)

Others have reported [no Internet access in the VM](https://gist.github.com/teohhanhui/042a395010d9946ceee14768736e3780?permalink_comment_id=5068762#gistcomment-5068762), probably related to this.

If I understand correctly, I think that's why `krunvm` set `nameserver 1.1.1.1` as the default:
https://github.com/containers/krunvm/blob/5494d84a66bee3b802a0392cf8d662158ac7287d/src/main.rs#L51

But that's also not a good solution as it'd break local domains and search domains among other things...

teohhanhui:
Weird... https://passt.top/passt/commit?id=bad252687271c0255f6a077f19cbc19aa0427f8d

slp:
> Which is bad, because that's my router's IP address... (And not what's sent as DNS servers via DHCP.)

That IP address points to `passt`'s DNS server, which will in turn use the host's DNS (as configured in the host's view of `/etc/resolv.conf`. This is how it's expected to work.

Which kind of problems are you having with DNS resolution? We would need to diagnose them first to be able to look for a solution.

teohhanhui:
> That IP address points to `passt`'s DNS server, which will in turn use the host's DNS (as configured in the host's view of `/etc/resolv.conf`. This is how it's expected to work.

Yeah, I realized that after looking around. It works for me anyway.

> Which kind of problems are you having with DNS resolution? We would need to diagnose them first to be able to look for a solution.

@aqrln Could you share what's in your `/etc/resolv.conf` both on the host and inside the VM?

@dylanchapell is somehow left only with the search domain inside the VM... Probably a `passt` bug? https://gist.github.com/teohhanhui/042a395010d9946ceee14768736e3780?permalink_comment_id=5073025#gistcomment-5073025

aqrln:
@teohhanhui 

> @aqrln Could you share what's in your /etc/resolv.conf both on the host and inside the VM?

Sorry I didn't have a chance to check yet (I needed to be on macOS since the beginning of the week), I saw your response to me in the gist as well. I'll get back to you asap next time I boot into Linux.

aqrln:
[...]

PR #64, opened 2024-09-26
Autoconfigure network without DHCP

slp:
Passt includes a DHCP server to allow guests to easily configure the network without being aware passt is on the other side. But we _are_ aware of that, so we can take advantage of it.

Instead of using DHCP, read the configuration output from passt, marshall it into some environment variables, pass it to the guest, and have krun-guest read it and apply it using rtnetlink.

By doing that, we reduce the startup time in half, from...

$ time krun /bin/false
(...)
real    0m4,301s

... to ...

$ time krun /bin/false
(...)
real    0m1,966s

### DIFF
--- a/crates/muvm/src/guest/net.rs
+++ b/crates/muvm/src/guest/net.rs
@@ -1,16 +1,26 @@
 use std::fs;
 use std::io::Write;
-use std::os::unix::process::ExitStatusExt as _;
-use std::process::Command;
+use std::net::{IpAddr, Ipv4Addr};
+use std::str::FromStr;
 
-use anyhow::{anyhow, Context, Result};
-use log::debug;
+use anyhow::{Context, Result};
+use futures_util::TryStreamExt;
+use rtnetlink::new_connection;
 use rustix::system::sethostname;
 
-use crate::utils::env::find_in_path;
-use crate::utils::fs::find_executable;
+use super::mount::place_etc;
 
-pub fn configure_network() -> Result<()> {
+pub struct NetworkConfig {
+    pub address: Option<String>,
+    pub mask: Option<String>,
+    pub router: Option<String>,
+    pub dns1: Option<String>,
+    pub dns2: Option<String>,
+    pub dns3: Option<String>,
+    pub search: Option<String>,
+}
+
+pub async fn configure_network(netconf: NetworkConfig) -> Result<()> {
     // Allow unprivileged users to use ping, as most distros do by default.
     {
         let mut file = fs::File::options()
@@ -33,63 +43,49 @@ pub fn configure_network() -> Result<()> {
         sethostname(hostname.as_bytes()).context("Failed to set hostname")?;
     }
 
-    let dhcpcd_path = find_in_path("dhcpcd").context("Failed to check existence of `dhcpcd`")?;
-    let dhcpcd_path = if let Some(dhcpcd_path) = dhcpcd_path {
-        Some(dhcpcd_path)
-    } else {
-        find_executable("/sbin/dhcpcd").context("Failed to check existence of `/sbin/dhcpcd`")?
-    };
-    if let Some(dhcpcd_path) = dhcpcd_path {
-        let output = Command::new(dhcpcd_path)
-            .args(["-M", "--nodev", "eth0"])
-            .output()
-            .context("Failed to execute `dhcpcd` as child process")?;
-        debug!(output:?; "dhcpcd output");
-        if !output.status.success() {
-            let err = if let Some(code) = output.status.code() {
-                anyhow!("`dhcpcd` process exited with status code: {code}")
-            } else {
-                anyhow!(
-                    "`dhcpcd` process terminated by signal: {}",
-                    output
-                        .status
-                        .signal()
-                        .expect("either one of status code or signal should be set")
-                )
-            };
-            Err(err)?;
+    let address = Ipv4Addr::from_str(&netconf.address.context("Missing MUVM_NETWORK_ADDRESS")?)?;
+    let mask = u32::from(Ipv4Addr::from_str(
+        &netconf.mask.context("Missing MUVM_NETWORK_MASK")?,
+    )?);
+    let prefix = (!mask).leading_zeros() as u8;
+    let router = netconf.router.context("Missing MUVM_NETWORK_ROUTER")?;
+    let router = Ipv4Addr::from_str(&router)?;
+
+    let (connection, handle, _) = new_connection().unwrap();
+    tokio::spawn(connection);
+    let mut links = handle.link().get().match_name("eth0".to_string()).execute();
+    if let Some(link) = links.try_next().await? {
+        handle
+            .address()
+            .add(link.header.index, IpAddr::V4(address), prefix)
+            .execute()
+            .await?;
+        handle.link().set(link.header.index).up().execute().await?
+    }
+    handle.route().add().v4().gateway(router).execute().await?;
+
+    // Only override resolv.conf is we have some values to put on it.
+    if netconf.dns1.is_some() || netconf.dns2.is_some() || netconf.dns3.is_some() {
+        place_etc("resolv.conf", None)?;
+        let mut resolv = fs::File::options()
+            .write(true)
+            .open("/etc/resolv.conf")
+            .context("Failed to open resolv.conf")?;
+
+        for ns in [netconf.dns1, netconf.dns2, netconf.dns3]
+            .into_iter()
+            .flatten()
+        {
+            resolv
+                .write_all(format!("nameserver {}\n", ns).as_bytes())
+                .context("Failed to write resolv.conf")?;
         }
 
-        return Ok(());
-    }
-
-    let dhclient_path =
-        find_in_path("dhclient").context("Failed to check existence of `dhclient`")?;
-    let dhclient_path = if let Some(dhclient_path) = dhclient_path {
-        Some(dhclient_path)
-    } else {
-        find_executable("/sbin/dhclient")
-            .context("Failed to check existence of `/sbin/dhclient`")?
-    };
-    let dhclient_path =
-        dhclient_path.ok_or_else(|| anyhow!("could not find required `dhcpcd` or `dhclient`"))?;
-    let output = Command::new(dhclient_path)
-        .output()
-        .context("Failed to execute `dhclient` as child process")?;
-    debug!(output:?; "dhclient output");
-    if !output.status.success() {
-        let err = if let Some(code) = output.status.code() {
-            anyhow!("`dhclient` process exited with status code: {code}")
-        } else {
-            anyhow!(
-                "`dhclient` process terminated by signal: {}",
-                output
-                    .status
-                    .signal()
-                    .expect("either one of status code or signal should be set")
-            )
-        };
-        Err(err)?;
+        if let Some(search) = netconf.search {
+            resolv
+                .write_all(format!("search {}\n", search).as_bytes())
+                .context("Failed to write resolv.conf")?;
+        }
     }
 
     Ok(())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,6 +82,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25bdb32cbbdce2b519a9cd7df3a678443100e265d5e25ca763b7572a5104f5f3"
 
 [[package]]
+name = "autocfg"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
+
+[[package]]
 name = "backtrace"
 version = "0.3.72"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -228,16 +234,93 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9fc0510504f03c51ada170672ac806f1f105a88aa97a5281117e1ddc3368e51a"
 
 [[package]]
-name = "futures-core"
-version = "0.3.30"
+name = "futures"
+version = "0.3.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfc6580bb841c5a68e9ef15c77ccc837b40a7504914d52e47b8b0e9bbda25a1d"
+checksum = "65bc07b1a8bc7c85c5f2e110c476c7389b4554ba72af57d8445ea63a576b0876"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-executor",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
+name = "futures-channel"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2dff15bf788c671c1934e366d07e30c1814a8ef514e1af724a602e8a2fbe1b10"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+]
+
+[[package]]
+name = "futures-core"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05f29059c0c2090612e8d742178b0580d2dc940c837851ad723096f87af6663e"
+
+[[package]]
+name = "futures-executor"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e28d1d997f585e54aebc3f97d39e72338912123a67330d723fdbb564d646c9f"
+dependencies = [
+ "futures-core",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
+name = "futures-io"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e5c1b78ca4aae1ac06c48a526a655760685149f0d465d21f37abfe57ce075c6"
+
+[[package]]
+name = "futures-macro"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "162ee34ebcb7c64a8abebc059ce0fee27c2262618d7b60ed8faf72fef13c3650"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
 
 [[package]]
 name = "futures-sink"
-version = "0.3.30"
+version = "0.3.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9fb8e00e87438d937621c1c6269e53f536c14d3fbd6a042bb24879e57d474fb5"
+checksum = "e575fab7d1e0dcb8d0c7bcf9a63ee213816ab51902e6d244a95819acacf1d4f7"
+
+[[package]]
+name = "futures-task"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f90f7dce0722e95104fcb095585910c0977252f286e354b5e3bd38902cd99988"
+
+[[package]]
+name = "futures-util"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9fa08315bb612088cc391249efdc3bc77536f16c91f6cf495e6fbe85b20a4a81"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-io",
+ "futures-macro",
+ "futures-sink",
+ "futures-task",
+ "memchr",
+ "pin-project-lite",
+ "pin-utils",
+ "slab",
+]
 
 [[package]]
 name = "getrandom"
@@ -372,9 +455,12 @@ dependencies = [
  "bpaf",
  "byteorder",
  "env_logger",
+ "futures-util",
  "krun-sys",
  "log",
- "nix",
+ "nix 0.28.0",
+ "regex",
+ "rtnetlink",
  "rustix",
  "serde",
  "serde_json",
@@ -382,6 +468,82 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "uuid",
+]
+
+[[package]]
+name = "netlink-packet-core"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72724faf704479d67b388da142b186f916188505e7e0b26719019c525882eda4"
+dependencies = [
+ "anyhow",
+ "byteorder",
+ "netlink-packet-utils",
+]
+
+[[package]]
+name = "netlink-packet-route"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74c171cd77b4ee8c7708da746ce392440cb7bcf618d122ec9ecc607b12938bf4"
+dependencies = [
+ "anyhow",
+ "byteorder",
+ "libc",
+ "log",
+ "netlink-packet-core",
+ "netlink-packet-utils",
+]
+
+[[package]]
+name = "netlink-packet-utils"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ede8a08c71ad5a95cdd0e4e52facd37190977039a4704eb82a283f713747d34"
+dependencies = [
+ "anyhow",
+ "byteorder",
+ "paste",
+ "thiserror",
+]
+
+[[package]]
+name = "netlink-proto"
+version = "0.11.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "86b33524dc0968bfad349684447bfce6db937a9ac3332a1fe60c0c5a5ce63f21"
+dependencies = [
+ "bytes",
+ "futures",
+ "log",
+ "netlink-packet-core",
+ "netlink-sys",
+ "thiserror",
+ "tokio",
+]
+
+[[package]]
+name = "netlink-sys"
+version = "0.8.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "416060d346fbaf1f23f9512963e3e878f1a78e707cb699ba9215761754244307"
+dependencies = [
+ "bytes",
+ "futures",
+ "libc",
+ "log",
+ "tokio",
+]
+
+[[package]]
+name = "nix"
+version = "0.27.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2eb04e9c688eff1c89d72b407f168cf79bb9e867a9d3323ed6c01519eb9cc053"
+dependencies = [
+ "bitflags",
+ "cfg-if",
+ "libc",
 ]
 
 [[package]]
@@ -426,10 +588,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
+
+[[package]]
 name = "pin-project-lite"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bda66fc9667c18cb2758a2ac84d1167245054bcf85d5d1aaa6923f45801bdd02"
+
+[[package]]
+name = "pin-utils"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "proc-macro2"
@@ -451,9 +625,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.10.4"
+version = "1.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c117dbdfde9c8308975b6a18d71f3f385c89461f7b3fb054288ecf2a2058ba4c"
+checksum = "38200e5ee88914975b69f657f0801b6f6dccafd44fd9326302a4aaeecfacb1d8"
 dependencies = [
  "aho-corasick",
  "memchr",
@@ -463,9 +637,9 @@ dependencies = [
 
 [[package]]
 name = "regex-automata"
-version = "0.4.6"
+version = "0.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86b83b8b9847f9bf95ef68afb0b8e6cdb80f498442f5179a29fad448fcc1eaea"
+checksum = "368758f23274712b504848e9d5a6f010445cc8b87a7cdb4d7cbee666c1288da3"
 dependencies = [
  "aho-corasick",
  "memchr",
@@ -474,9 +648,27 @@ dependencies = [
 
 [[package]]
 name = "regex-syntax"
-version = "0.8.3"
+version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "adad44e29e4c806119491a7f06f03de4d1af22c3a680dd47f1e6e179439d1f56"
+checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
+
+[[package]]
+name = "rtnetlink"
+version = "0.14.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b684475344d8df1859ddb2d395dd3dac4f8f3422a1aa0725993cb375fc5caba5"
+dependencies = [
+ "futures",
+ "log",
+ "netlink-packet-core",
+ "netlink-packet-route",
+ "netlink-packet-utils",
+ "netlink-proto",
+ "netlink-sys",
+ "nix 0.27.1",
+ "thiserror",
+ "tokio",
+]
 
 [[package]]
 name = "rustc-demangle"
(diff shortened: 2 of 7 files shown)


crates/krun/Cargo.toml
```diff
@@ -13,10 +13,13 @@ anyhow = { version = "1.0.82", default-features = false, features = ["std"] }
 bpaf = { version = "0.9.11", default-features = false, features = [] }
 byteorder = { version = "1.5.0", default-features = false, features = ["std"] }
 env_logger = { version = "0.11.3", default-features = false, features = ["auto-color", "humantime", "unstable-kv"] }
+futures-util = "0.3.30"
```
teohhanhui, line 16: Could we add `default-features = false, features = [...]` for all?

crates/krun/src/guest/bin/krun-guest.rs
```diff
@@ -15,7 +15,8 @@ use krun::utils::env::find_in_path;
 use log::debug;
 use rustix::process::{getrlimit, setrlimit, Resource};
 
-fn main() -> Result<()> {
+#[tokio::main]
```
teohhanhui, line 18: Some of our safety assumptions would no longer hold, e.g.

https://github.com/slp/krun/blob/44417f6e802a934f9ca164ace647f99da009546d/crates/krun/src/guest/user.rs#L21-L23

slp: Using tokio is a requirement of the rtnetlink crate. I think we can avoid the safety issues with something like this:
```
    let rt = tokio::runtime::Runtime::new().unwrap();
    rt.block_on(async {
        if let Err(err) = configure_network().await {
            eprintln!("Failed to configure network, continuing without it: {err}");
        }
    });
```

teohhanhui: Could we use tokio's current-thread scheduler?

crates/krun/Cargo.toml
slp: Done, thanks!

crates/krun/src/guest/bin/krun-guest.rs
slp: If we aren't planning on using tokio anywhere else, I think I feel more confident isolating it's use to `configure_network`.

crates/krun/src/guest/bin/krun-guest.rs
```diff
@@ -42,7 +42,12 @@ fn main() -> Result<()> {
 
     setup_fex()?;
 
-    configure_network()?;
+    let rt = tokio::runtime::Runtime::new().unwrap();
+    rt.block_on(async {
```
teohhanhui, line 46: If I understand the docs correctly, this will NOT have the desired effect of ensuring the safety of the `env::set_var` calls.

> This runs the given future on the current thread, blocking until it is complete, and yielding its resolved result. Any tasks or timers which the future spawns internally will be executed on the runtime.

> When the multi thread scheduler is used this will allow futures to run within the io driver and timer context of the overall runtime.
>
> Any spawned tasks will continue running after block_on returns.

https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on

slp: I think you're right. Another approach could be, since I don't think any of the other tasks `krun-guest` puts in motion depends on the network, simply move the network configuration to `krun-server` which already uses tokio and doesn't use `env::set_var`.

crates/muvm/src/guest/net.rs
```diff
@@ -33,64 +37,28 @@ pub fn configure_network() -> Result<()> {
...
+    if let Some(link) = links.try_next().await? {
+        handle
+            .address()
+            .add(link.header.index, IpAddr::V4(address), prefix)
+            .execute()
+            .await?;
+        handle.link().set(link.header.index).up().execute().await?
     }
+    handle.route().add().v4().gateway(router).execute().await?;
+    fs::write("/etc/resolv.conf", format!("nameserver {}", router))
```
WhatAmISupposedToPutHere, line 60: You are assuming that the default gateway and the dns server are the same. While true on most residential networks, it is not a correct assumption in general, and even there, some of us very intentionally override the ISP dns server to something else.

teohhanhui: This is inside the VM, so `passt` has already done its thing. In our current usage, `/etc/resolv.conf` inside the VM would contain a `nameserver` entry pointing to the host's default gateway...

https://man.archlinux.org/man/passt.1.en#Handling_of_traffic_with_local_destination_and_source_addresses

WhatAmISupposedToPutHere: > /etc/resolv.conf inside the VM would contain a nameserver entry pointing to the host's default gateway.

That is the problem that i am pointing out, in general nameserver != default gateway.

teohhanhui: More correct explanation here: https://github.com/AsahiLinux/muvm/issues/17#issuecomment-2160748250

teohhanhui: @WhatAmISupposedToPutHere Uhh, but I think I get what you mean now... The code here is only correct in the case of `passt` doing such DNS forwarding... If the resolver on the host is not a loopback address, this code is wrong.

slp: Good catch. I've updated the PR to actually pick up all the DNS configuration from passt, which was the right thing to do.

sbrivio-rh:
> But sure, you'll need to convey DNS information somehow.

Actually, even for that, you could hardcode things in a way similar to what Podman does with pasta(1): a single link-local address in `/etc/resolv.conf` in the container (here, guest) and tell passt to map that (`--dns-forward`) to whatever resolver you have on the host. On the other hand:

> You could happily use link-local addresses (even for IPv4, say, 169.254.1.1)

this way, you would lose network transparency (making the guest look like the host in terms of addresses and routes), which might be important for some applications. I wanted to try out the tricks I suggested above to make DHCP fast, but I'm currently hitting:

```
   Compiling devices v0.1.0 (/home/sbrivio/libkrun/src/devices)
error: couldn't read `src/devices/src/virtio/fs/linux/../../../../../../init/init`: No such file or directory (os error 2)
  --> src/devices/src/virtio/fs/linux/passthrough.rs:33:29
   |
33 | static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `devices` (lib) due to 1 previous error
```

while trying to build libkrun. I might get back to it in a few days unless there's an obvious solution.

sbrivio-rh:
From r/Showerthoughts: netlink over vsock

slp:
> this way, you would lose network transparency (making the guest look like the host in terms of addresses and routes), which might be important for some applications. I wanted to try out the tricks I suggested above to make DHCP fast, but I'm currently hitting:
> 
> ```
>    Compiling devices v0.1.0 (/home/sbrivio/libkrun/src/devices)
> error: couldn't read `src/devices/src/virtio/fs/linux/../../../../../../init/init`: No such file or directory (os error 2)
>   --> src/devices/src/virtio/fs/linux/passthrough.rs:33:29
>    |
> 33 | static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
>    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>    |
>    = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)
> 
> error: could not compile `devices` (lib) due to 1 previous error
> ```
> 
> while trying to build libkrun. I might get back to it in a few days unless there's an obvious solution.

You're probably running `cargo build ...` instead of make. Run `make init/init` once, and then you can rely solely con cargo.

sbrivio-rh:
> You're probably running `cargo build ...` instead of make.

Oops, of course.

> Run `make init/init` once, and then you can rely solely con cargo.

Right, it works, thanks! I'll try things out in a bit.

sbrivio-rh:
> Right, it works, thanks! I'll try things out in a bit.

...currently driving me mad (yes, I have libkrun and libkrunfw installed):
```
  = note: /usr/bin/ld: /home/sbrivio/muvm/target/debug/deps/muvm-dfc2b43586ae1702.73tss62c465d1mti0x3tsc1b1.rcgu.o: in function `muvm::add_ro_disk':
          /home/sbrivio/muvm/crates/muvm/src/bin/muvm.rs:51:(.text._ZN4muvm11add_ro_disk17hc95bc1821f59bb8eE+0x362): undefined reference to `krun_add_disk'
          collect2: error: ld returned 1 exit status
          
  = note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib)

error: could not compile `muvm` (bin "muvm") due to 1 previous error
```

teohhanhui:
@sbrivio-rh I think it's a cache problem. Try doing `cargo clean` and try again?

sbrivio-rh:
> EDIT: Ohhhhh... [#84 (comment)](https://github.com/AsahiLinux/muvm/issues/84#issuecomment-2407426931)

Thanks, I would have never found that!

sbrivio-rh:
Almost there...

<details>
```
$ strace -f ./muvm true

[...]

[pid 4013940] execve("/usr/local/bin/passt", ["passt", "-q", "-f", "-t", "3334:3334", "--fd", "6"], 0x7ffd61f5fc80 /* 23 vars */) = -1 ENOENT (No such file or directory)
[pid 4013940] execve("/usr/bin/passt", ["passt", "-q", "-f", "-t", "3334:3334", "--fd", "6"], 0x7ffd61f5fc80 /* 23 vars */ <unfinished ...>
[pid 4013939] <... clone3 resumed>)     = 4013940
[pid 4013939] munmap(0x7f1bfb842000, 36864) = 0
[pid 4013939] rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
[pid 4013939] fcntl(3, F_GETFD <unfinished ...>
[pid 4013940] <... execve resumed>)     = 0
[pid 4013939] <... fcntl resumed>)      = 0x1 (flags FD_CLOEXEC)
[pid 4013939] close(3)                  = 0
[pid 4013939] write(2, "Error: ", 7 <unfinished ...>
Error: [pid 4013940] brk(NULL <unfinished ...>
[pid 4013939] <... write resumed>)      = 7
[pid 4013940] <... brk resumed>)        = 0x563b6c492000
[pid 4013939] write(2, "Failed to configure net mode", 28Failed to configure net mode) = 28
[pid 4013939] write(2, "\n\nCaused by:", 12 <unfinished ...>


Caused by:[pid 4013940] mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0 <unfinished ...>
[pid 4013939] <... write resumed>)      = 12
[pid 4013940] <... mmap resumed>)       = 0x7f1605c01000
[pid 4013939] write(2, "\n", 1
)         = 1
[pid 4013940] access("/etc/ld.so.preload", R_OK <unfinished ...>
[pid 4013939] write(2, "    ", 4     <unfinished ...>
[pid 4013940] <... access resumed>)     = -1 ENOENT (No such file or directory)
[pid 4013939] <... write resumed>)      = 4
[pid 4013939] write(2, "Operation not supported", 23 <unfinished ...>
Operation not supported[pid 4013940] openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC <unfinished ...>
[pid 4013939] <... write resumed>)      = 23
[pid 4013939] write(2, " (os error ", 11 <unfinished ...>
 (os error [pid 4013940] <... openat resumed>)     = 3
[pid 4013939] <... write resumed>)      = 11
[pid 4013940] fstat(3,  <unfinished ...>
[pid 4013939] write(2, "95", 295 <unfinished ...>
```

...but I can't see anything returning -95 / `EOPNOTSUPP`...
</details>

sbrivio-rh:
So, at least on my setup, this whole delay is actually caused by dhcpcd(8):

```
$ ./passt -f -p /tmp/muvm.pcap
```

```
$ time ./muvm --passt-socket=/tmp/passt_3.socket /bin/false 2>/dev/null

real	0m6.445s
user	0m0.916s
sys	0m0.549s
```

```
$ tshark -r /tmp/muvm.pcap 
    1   0.000000      0.0.0.0 → 255.255.255.255 DHCP 342 DHCP Discover - Transaction ID 0x356a31d7
    2   0.000148 88.198.0.161 → 88.198.0.164 DHCP 342 DHCP Offer    - Transaction ID 0x356a31d7
    3   0.000513      0.0.0.0 → 255.255.255.255 DHCP 342 DHCP Request  - Transaction ID 0x356a31d7
    4   0.000542 88.198.0.161 → 88.198.0.164 DHCP 342 DHCP ACK      - Transaction ID 0x356a31d7
    5   0.023915 5a:94:ef:e4:0c:ee → Broadcast    ARP 42 Who has 88.198.0.164? (ARP Probe)
    6   1.448584 5a:94:ef:e4:0c:ee → Broadcast    ARP 42 Who has 88.198.0.164? (ARP Probe)
    7   2.828025 5a:94:ef:e4:0c:ee → Broadcast    ARP 42 Who has 88.198.0.164? (ARP Probe)
    8   4.830970 5a:94:ef:e4:0c:ee → Broadcast    ARP 42 ARP Announcement for 88.198.0.164
```

because, by default (without `--noarp`), it ARP-probes the address we just assigned. If I do this:

```
diff --git a/crates/muvm/src/guest/net.rs b/crates/muvm/src/guest/net.rs
index b47f2e1..aa74ce0 100644
--- a/crates/muvm/src/guest/net.rs
+++ b/crates/muvm/src/guest/net.rs
@@ -41,7 +41,7 @@ pub fn configure_network() -> Result<()> {
     };
     if let Some(dhcpcd_path) = dhcpcd_path {
         let output = Command::new(dhcpcd_path)
-            .args(["-M", "--nodev", "eth0"])
+            .args(["-M", "--nodev", "eth0", "--noarp"])
             .output()
             .context("Failed to execute `dhcpcd` as child process")?;
         debug!(output:?; "dhcpcd output");
```

I get:

```
$ time ./muvm --passt-socket=/tmp/passt_3.socket /bin/false 2>/dev/null

real	0m1.404s
user	0m0.899s
sys	0m0.583s
```

For some reason I couldn't get IPv6 working in the guest so I didn't try that (at least, not yet).

sbrivio-rh:
> For some reason I couldn't get IPv6 working in the guest so I didn't try that (at least, not yet).

Oh, it's simply disabled in the kernel configuration from libkrunfw (config-libkrunfw_x86_64):
```
# CONFIG_IPV6 is not set
```
and this is there starting from the very beginning, libkrunfw commit 443c03426a73. Is there a specific reason for it? Maybe because TSI didn't/doesn't support that? Or you don't need it at all in libkrun/muvm? Does it make sense that I spend a moment trying to "fix" that? @slp?

slp:
> and this is there starting from the very beginning, libkrunfw commit 443c03426a73. Is there a specific reason for it? Maybe because TSI didn't/doesn't support that? Or you don't need it at all in libkrun/muvm? Does it make sense that I spend a moment trying to "fix" that? @slp?

That was disabled as part of the work to slim down the guest kernel. I'm okay with enabling it if there's a user demand for the feature, but so far I haven't heard of anyone requesting it.

teohhanhui:
It just feels *ethically* wrong to be contributing to holding back IPv6 adoption. 🙈

asahilina:
Oof, libkrunfw doesn't have `CONFIG_IPV6`? Yeah, we really should fix that... even if things work with IPv4 only for most people, it's not sustainable to not have any kind of v6 networking within the VM. Even if v4 only works, IPv4 infrastructure is increasingly going through CGNAT while IPv6 infrastructure is not, and that means IPv6 is often faster where it is available (that is the case for me most of the time at home).

sbrivio-rh:
> That was disabled as part of the work to slim down the guest kernel. I'm okay with enabling it if there's a user demand for the feature, but so far I haven't heard of anyone requesting it.

There's one problem with that and passt: if passt doesn't find an IPv4 interface on the host, it disables IPv4, and only smoke signals remain at that point.

This is by default at least: you can still enable IPv4 even on an IPv6-only host, but then you need to give explicit addresses and routes.

I can try to enable `CONFIG_IPV6` in libkrunfw and see what happens. I'll give it a try at some point this week.

sbrivio-rh:
> I can try to enable `CONFIG_IPV6` in libkrunfw and see what happens. I'll give it a try at some point this week.

I meant _this_ week. IPv6 enabled in libkrunfw and everything, but at this point I would really need to get a shell in muvm to check addresses and routes, and I actually never got that far:
```
$ ./muvm --passt-socket=/tmp/passt_1.socket /bin/false
Using default interface naming scheme 'v255'.
/etc/udev/rules.d/60-scheduler.rules:1 The line has no effect, ignoring.
Error: Failed to set up user, bailing out

Caused by:
    0: Failed to read directory `/dev/dri`
    1: No such file or directory (os error 2)
```

I don't have DRI/DRM support on this system (I don't have a graphical environment or a display adapter either). How do I work around that?

asahilina:
That's a bug... let me fix it.

sbrivio-rh:
> Now looking into IPv6 support via passt.

Yeah, it works almost out of the box, with a simple `CONFIG_IPV6=y` in libkrunfw. I just need to bring up the link before `dhcpcd` starts. Now trying with a few combinations (also with `dhclient`) and checking timing...

slp:
FWIW, the latest libkrunfw (4.5.1) has CONFIG_IPV6 enabled.

sbrivio-rh:
> FWIW, the latest libkrunfw (4.5.1) has CONFIG_IPV6 enabled.

Yes, I noticed, thanks. If I simply:

```
diff --git a/crates/muvm/src/guest/net.rs b/crates/muvm/src/guest/net.rs
index b47f2e1..96c4689 100644
--- a/crates/muvm/src/guest/net.rs
+++ b/crates/muvm/src/guest/net.rs
@@ -33,6 +33,8 @@ pub fn configure_network() -> Result<()> {
         sethostname(hostname.as_bytes()).context("Failed to set hostname")?;
     }
 
+    let output = Command::new("/sbin/ip").args(["link", "set", "dev", "eth0", "up"]).output();
+
     let dhcpcd_path = find_in_path("dhcpcd").context("Failed to check existence of `dhcpcd`")?;
     let dhcpcd_path = if let Some(dhcpcd_path) = dhcpcd_path {
         Some(dhcpcd_path)
```

...I get IPv6 addresses and routes now, but we need to disable DAD on the address (which means we need to disable neighbour advertisements on the interface, bring it up, set `IFA_F_NODAD` on the address, re-enable advertisements), otherwise we add two seconds there. That needs a few netlink messages, doing everything with ip(8) takes ages.

For IPv4, I'm trying out DHCP rapid commit (RFC 4039) instead.

This might take a bit, but I think we can have IPv4 and IPv6 connectivity set up fast (a couple of milliseconds) and relatively cleanly.

sbrivio-rh:
> This might take a bit, but I think we can have IPv4 and IPv6 connectivity set up fast (a couple of milliseconds) and relatively cleanly.

I guess I can claim success:

```
$ time ./target/debug/muvm --passt-socket=/tmp/passt_3.socket -- /bin/false
DHCP client took 340.27µs
Network configuration took 1.5503ms
Using default interface naming scheme 'v255'.
"/bin/false" process exited with status code: 1

real	0m1.196s
user	0m0.747s
sys	0m0.502s
```

with both IPv4 (DHCP) and IPv6 (SLAAC) up and running, with the net.rs below plus some [changes](https://archives.passt.top/passt-dev/20241125090008.132338-1-sbrivio@redhat.com/) in passt implementing option 80 and honouring the "broadcast" DHCP flag (not merged yet, but it shouldn't take long before we get it in packages).

A short summary of the proposed implementation:
* External DHCP Clients Considered Harmful
* instead of just triggering network configuration and then running away, wait until it completes, to guarantee that we can (correctly) run whatever application directly from the command line. Perhaps this is not so important for the current (main?) usage of muvm, but I have some secret plans about using this thing for kselftests which are not secret anymore
* for IPv4-only: use DHCP, then poll on the `IFA_F_TENTATIVE` flag of the link-local IPv6 address (`CONFIG_IPV6=y` in libkrunfw is enough to get one) to see if we can expect a global unicast address coming from passt via SLAAC
* for IPv6-only: send the DHCP request with a short (?) timeout (100ms) to keep that case reasonably fast
* both IPv4 and IPv6 (passt's default if both are available on the host): start SLAAC, take care of DHCP, then wait until SLAAC completed

<details><summary>Proposed implementation (still with "profiling" prints):</summary>

```

use std::fs;
use std::io::Write;
use std::net::{UdpSocket, Ipv4Addr};
use std::time::Instant;
use std::time::Duration;

use anyhow::{Context, Result};
use rustix::system::sethostname;

use neli::{
    consts::{
        nl::NlmF,
        rtnl::{Arphrd, Ifa, IfaF, Iff, Rta, RtAddrFamily, Rtm, RtmF, Rtn,
               Rtprot, RtScope, RtTable},
        socket::NlFamily,
    },
    nl::{NlPayload, Nlmsghdr},
    router::synchronous::{NlRouter, NlRouterReceiverHandle},
    rtnl::{Ifaddrmsg, IfaddrmsgBuilder, Ifinfomsg, IfinfomsgBuilder,
           RtattrBuilder, Rtmsg, RtmsgBuilder},
    utils::Groups,
    types::RtBuffer,
};

/// Set interface flags for eth0 (interface index 2) with a given mask
fn flags_eth0(rtnl: &NlRouter, mask: Iff, set: Iff) -> Result<()> {
    let ifinfomsg = IfinfomsgBuilder::default()
        .ifi_family(RtAddrFamily::Unspecified)
        .ifi_type(Arphrd::Ether).ifi_index(2)
        .ifi_change(mask).ifi_flags(set)
        .build()?;

    let _: NlRouterReceiverHandle<Rtm, Ifinfomsg> =
        rtnl.send(Rtm::Newlink, NlmF::REQUEST, NlPayload::Payload(ifinfomsg))?;

    Ok(())
}

/// Add or delete IPv4 routes for eth0 (interface index 2)
fn route4_eth0(rtnl: &NlRouter, what: Rtm, gw: Ipv4Addr) -> Result<()> {
    let rtmsg = RtmsgBuilder::default()
        .rtm_family(RtAddrFamily::Inet)
        .rtm_dst_len(0).rtm_src_len(0).rtm_tos(0)
        .rtm_table(RtTable::Main).rtm_protocol(Rtprot::Boot)
        .rtm_scope(RtScope::Universe).rtm_type(Rtn::Unicast)
        .rtm_flags(RtmF::empty())
        .rtattrs(RtBuffer::from_iter([
            RtattrBuilder::default()
                .rta_type(Rta::Oif)
                .rta_payload(2)
                .build()?,
            RtattrBuilder::default()
                .rta_type(Rta::Dst)
                .rta_payload(Ipv4Addr::UNSPECIFIED.octets().to_vec())
                .build()?,
            RtattrBuilder::default()
                .rta_type(Rta::Gateway)
                .rta_payload(gw.octets().to_vec())
                .build()?
        ]))
        .build()?;

    let _: NlRouterReceiverHandle<Rtm, Rtmsg> =
        rtnl.send(what, NlmF::CREATE | NlmF::REQUEST,
                  NlPayload::Payload(rtmsg))?;

    Ok(())
}

/// Add or delete IPv4 addresses for eth0 (interface index 2)
fn addr4_eth0(rtnl: &NlRouter, what: Rtm, addr: Ipv4Addr, prefix_len: u8)
                       -> Result<()> {
    let ifaddrmsg = IfaddrmsgBuilder::default()
        .ifa_family(RtAddrFamily::Inet)
        .ifa_prefixlen(prefix_len)
        .ifa_scope(RtScope::Universe)
        .ifa_index(2)
        .rtattrs(RtBuffer::from_iter([
            RtattrBuilder::default()
                .rta_type(Ifa::Local)
                .rta_payload(addr.octets().to_vec())
                .build()?,
            RtattrBuilder::default()
                .rta_type(Ifa::Address)
                .rta_payload(addr.octets().to_vec())
                .build()?,
        ]))
        .build()?;

    let _: NlRouterReceiverHandle<Rtm, Ifaddrmsg> =
        rtnl.send(what, NlmF::CREATE | NlmF::REQUEST,
                  NlPayload::Payload(ifaddrmsg))?;

    Ok(())
}

pub fn configure_network() -> Result<()> {

    // Allow unprivileged users to use ping, as most distros do by default.
    {
        let mut file = fs::File::options()
            .write(true)
            .open("/proc/sys/net/ipv4/ping_group_range")
            .context("Failed to open ipv4/ping_group_range for writing")?;

        file.write_all(format!("{} {}", 0, 2147483647).as_bytes())
            .context("Failed to extend ping group range")?;
    }

    {
        let hostname =
            fs::read_to_string("/etc/hostname").unwrap_or("placeholder-hostname".to_string());
        let hostname = if let Some((hostname, _)) = hostname.split_once('\n') {
            hostname.to_owned()
        } else {
            hostname
        };
        sethostname(hostname.as_bytes()).context("Failed to set hostname")?;
    }

    let start = Instant::now();
    let (rtnl, _) = NlRouter::connect(NlFamily::Route, None, Groups::empty())?;
    rtnl.enable_strict_checking(true)?;

    // Disable neighbour solicitations (dodge DAD), bring up link to start SLAAC
    {
        // IFF_NOARP | IFF_UP in one shot delays router solicitations, avoid it
        flags_eth0(&rtnl, Iff::NOARP, Iff::NOARP)?;
        flags_eth0(&rtnl, Iff::UP, Iff::UP)?;
    }

    let start_dhcp = Instant::now();
    // Configure IPv4 using DHCP with Rapid Commit (DISCOVER -> ACK)
    {
        // Temporary link-local address and route avoid the need for raw sockets
        route4_eth0(&rtnl, Rtm::Newroute, Ipv4Addr::UNSPECIFIED)?;
        addr4_eth0(&rtnl, Rtm::Newaddr, Ipv4Addr::new(169, 254, 1, 1), 16)?;

        // Send request (DHCPDISCOVER)
        let socket = UdpSocket::bind("0.0.0.0:68").expect("Failed to bind");
        let mut buf = [0; 576 /* RFC 2131, Section 2 */ ];

        const REQUEST: [u8; 300 /* From RFC 951: >= 60 B of options */ ] = [
            1 /* REQUEST */, 0x1 /* Ethernet */, 6 /* hlen */, 0 /* Hops */,
            1, 2, 3, 4 /* XID */, 0, 0 /* Seconds */, 0x80, 0x0 /* Flags */,
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* All-zero (four) addresses */
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* 16B HW address: who cares */
            /* 32 bytes per row: 64B 'sname', plus 128B 'file' (RFC 1531) */
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
            0x63, 0x82, 0x53, 0x63, /* DHCP (magic) cookie, then options: */
            53, 1, 1 /* DISCOVER */, 80, 0 /* Rapid commit */, 0xff, // Done
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
            0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0 /* 54B paaaadding */
        ];

        socket.set_broadcast(true)?;
        socket.send_to(&REQUEST, "255.255.255.255:67")?;

        // Keep IPv6-only fast
        let _ = socket.set_read_timeout(Some(Duration::from_millis(100)));

        // Get and process response (DHCPACK) if any
        if let Ok((len, _)) = socket.recv_from(&mut buf) {
            let msg = &mut buf[..len];

            let addr = Ipv4Addr::new(msg[16], msg[17], msg[18], msg[19]);
            let mut netmask = Ipv4Addr::UNSPECIFIED;
            let mut router = Ipv4Addr::UNSPECIFIED;
            let mut p: usize = 240;

            while p < len {
                let o = msg[p];
                let mut l: u8;
                l = msg[p + 1];

                if o == 1 {           // Option 1: Subnet Mask
                    netmask = Ipv4Addr::new(msg[p + 2], msg[p + 3],
                                            msg[p + 4], msg[p + 5]);
                } else if o == 3 {    // Option 3: Router
                    router =  Ipv4Addr::new(msg[p + 2], msg[p + 3],
                                            msg[p + 4], msg[p + 5]);
                } else if o == 0xff { // Option 255: End (of options)
                    break;
                }

                l += 2; // Length doesn't include code and length field itself
                p += l as usize;
            }

            let prefix_len : u8 = netmask.to_bits().leading_ones() as u8;

            // Drop temporary address and route, configure what we got instead
            route4_eth0(&rtnl, Rtm::Delroute, Ipv4Addr::UNSPECIFIED)?;
            addr4_eth0(&rtnl, Rtm::Deladdr, Ipv4Addr::new(169, 254, 1, 1), 16)?;

            addr4_eth0(&rtnl, Rtm::Newaddr, addr, prefix_len)?;
            route4_eth0(&rtnl, Rtm::Newroute, router)?;
        } else {
            // Clean up: we're clearly too cool for IPv4
            route4_eth0(&rtnl, Rtm::Delroute, Ipv4Addr::UNSPECIFIED)?;
            addr4_eth0(&rtnl, Rtm::Deladdr, Ipv4Addr::new(169, 254, 1, 1), 16)?;
        }
    }
    let elapsed = start_dhcp.elapsed();
    println!("DHCP client took {:?}", elapsed);

    // Wait for SLAAC to complete or fail: we're done only once network is ready
    {
        let mut global_seen = false;
        let mut global_wait = true;
        let mut ll_seen = false;

        // Busy-netlink-loop until we see a link-local address, and a global
        // unicast address as long as we might expect one (see below)
        while !ll_seen || (global_wait && !global_seen) {
            let ifaddrmsg = IfaddrmsgBuilder::default()
                .ifa_family(RtAddrFamily::Inet6)
                .ifa_prefixlen(0)
                .ifa_scope(RtScope::Universe)
                .ifa_index(2)
                .build()?;

            let recv = rtnl.send(Rtm::Getaddr, NlmF::ROOT,
                                 NlPayload::Payload(ifaddrmsg))?;

            for response in recv {
                let header: Nlmsghdr<Rtm, Ifaddrmsg> = response?;
                if let NlPayload::Payload(p) = header.nl_payload() {
                    if p.ifa_scope() == &RtScope::Link {
                        // A non-tentative link-local address implies we sent a
                        // router solicitation that didn't get any response
                        // (IPv4-only)? Stop waiting for the router in that case
                        if *p.ifa_flags() & IfaF::TENTATIVE != IfaF::TENTATIVE {
                            global_wait = false;
                        }

                        ll_seen = true;
                    } else if p.ifa_scope() == &RtScope::Universe {
                        global_seen = true;
                    }
                }
            }
        }
    }

    // Re-enable neighbour solicitations and ARP requests
    {
        flags_eth0(&rtnl, Iff::NOARP, Iff::empty())?;
    }

    let elapsed = start.elapsed();
    println!("Network configuration took {:?}", elapsed);

    Ok(())
}
```
</details>

<details><summary>Both link-local and global connectivity work right away</summary>

```

$ time ./target/debug/muvm --passt-socket=/tmp/passt_3.socket -- ping -c1 2a01:4f8:222:904::2
DHCP client took 571.66µs
Network configuration took 1.83646ms
Using default interface naming scheme 'v255'.
PING 2a01:4f8:222:904::2 (2a01:4f8:222:904::2) 56 data bytes
64 bytes from 2a01:4f8:222:904::2: icmp_seq=1 ttl=255 time=0.256 ms

--- 2a01:4f8:222:904::2 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.256/0.256/0.256/0.000 ms

real	0m1.191s
user	0m0.746s
sys	0m0.493s

$ time ./target/debug/muvm --passt-socket=/tmp/passt_3.socket -- ping -c1 fe80::1%eth0
DHCP client took 396.18µs
Network configuration took 1.79467ms
Using default interface naming scheme 'v255'.
PING fe80::1%eth0 (fe80::1%eth0) 56 data bytes
64 bytes from fe80::1%eth0: icmp_seq=1 ttl=255 time=0.234 ms

--- fe80::1%eth0 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.234/0.234/0.234/0.000 ms

real	0m1.199s
user	0m0.713s
sys	0m0.528s
```
</details>

<details><summary>IPv6-only operation (`passt -6`)</summary>

```

$ time ./target/debug/muvm --passt-socket=/tmp/passt_3.socket -- /bin/false
Using default interface naming scheme 'v255'.
/etc/udev/rules.d/60-scheduler.rules:1 The line has no effect, ignoring.
DHCP client took 101.189115ms
Network configuration took 102.811555ms
"/bin/false" process exited with status code: 1

real	0m1.288s
user	0m0.811s
sys	0m0.626s
```
</details>

<details><summary>IPv4-only operation (`passt -4`)</summary>

```

$ time ./target/debug/muvm --passt-socket=/tmp/passt_3.socket -- /bin/false
DHCP client took 377.87µs
Network configuration took 1.77258ms
"/bin/false" process exited with status code: 1

real	0m1.158s
user	0m0.689s
sys	0m0.508s
```
</details>

What do you think? Should I submit this as a separate change?

slp:
> What do you think? Should I submit this as a separate change?

This looks great, thank you @sbrivio-rh !

Yes please, submit this as a new PR that will supersede this one. I guess we need to wait until the new passt lands in Fedora?

sbrivio-rh:
> Yes please, submit this as a new PR that will supersede this one.

Okay! I'll do that in a bit.

> I guess we need to wait until the new passt lands in Fedora?

Yes, and that we merge the series, too. But after that part, I'm usually fast.

slp:
This one got deprecated by #111

PR #65, opened 2024-09-27
Pass --direct-scale and --log-level to sommelier

slp:
With HiDPI screens, some games running in sommelier may scale way beyond the actual screen size. Passing --direct-scale to sommelier fixes this.

While there, also pass --log-level=3 to reduce sommelier's verbosity which was increased in the latest versions.


alyssarosenzweig (approved):
Sane.

PR #66, opened 2024-09-27
guest: allow unprivileged users to use ping

slp:
By default, the kernel doesn't allow unprivileged users to create ICMP Echo sockets, which can lead to confusing messages like this one:

$ ping 1.1.1.1
ping: socktype: SOCK_DGRAM
ping: socket: Address family not supported by protocol

Most distros adjust "ipv4/ping_group_range" to allow unprivileged users to use "ping" without relying on setuid, so do the same for krun guests.


alyssarosenzweig (approved):
nice little QoL improvement

PR #67, opened 2024-09-27
Minor clippy fixes and add a minimal workflow checking format and clippy

slp:


.github/workflows/code_quality.yml
```diff
@@ -0,0 +1,33 @@
+name: Code Quality (rustfmt and clippy)
```
teohhanhui, line 1:
```suggestion
name: Code Quality
```

The UI shows the job names, so there's no need to repeat them here.

.github/workflows/code_quality.yml
```diff
@@ -0,0 +1,36 @@
+name: Code Quality
+on: [pull_request, create]
```
teohhanhui, line 2:
```suggestion
on:
  push:
    branches:
      - main
  pull_request:
  create:
```

We should at least run this on push to `main`, i.e. after any PR is merged.

.github/workflows/code_quality.yml
slp: done

alyssarosenzweig (approved):

PR #68, opened 2024-09-27
Allow fex-emu config to fail

alyssarosenzweig:
These dirs may not exist depending how fex-emu was installed.

### DIFF
--- a/crates/krun/src/guest/mount.rs
+++ b/crates/krun/src/guest/mount.rs
@@ -80,11 +80,13 @@ fn mount_fex_rootfs() -> Result<()> {
     // and a Config.json telling FEX to use FUSE. Neither should be visible to the guest. Instead,
     // we want to replace the folders and tell FEX to use our mounted rootfs
     for base in ["/usr/share/fex-emu", "/usr/local/share/fex-emu"] {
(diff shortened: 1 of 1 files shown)


crates/krun/src/guest/mount.rs
```diff
@@ -80,11 +80,13 @@ fn mount_fex_rootfs() -> Result<()> {
     // and a Config.json telling FEX to use FUSE. Neither should be visible to the guest. Instead,
     // we want to replace the folders and tell FEX to use our mounted rootfs
     for base in ["/usr/share/fex-emu", "/usr/local/share/fex-emu"] {
-        let json = format!("{{\"Config\":{{\"RootFS\":\"{dir_rootfs}\"}}}}\n");
-        let path = base.to_string() + "/Config.json";
+        if Path::new(base).exists() {
```
teohhanhui, line 83:
```suggestion
        if fs::exists(base) {
```

https://doc.rust-lang.org/stable/std/fs/fn.exists.html

alyssarosenzweig: the doc mentioned TOCTOU bugs and then I realized there's a better way. fixed.

slp (approved):
LGTM

PR #69, opened 2024-09-28
krun: set the vRAM size for the microVM

slp:
Starting v1.9.5, libkrun allows configuring the size of the SHM window
for virtio-gpu, which acts as a pseudo-vram (previously it was hardcoded
to a conservative 8GB).

Let's use that feature to configure the vRAM of the microVM to a
reasonable value. Since we expect userspace drivers in the guest to be
conservative for us, telling appliations that vRAM is just a percentage
of the guest's RAM, let's just set vRAM to be the same as the total
amount of RAM in the system, leaving some room for potential
fragmentation.

Also, introduce the "--vram" command flag to allow users to override the
default value.

While there, also remove the 32 GB limit, since libkrun does support
guests larger than that.

### DIFF
--- a/crates/krun/src/bin/krun.rs
+++ b/crates/krun/src/bin/krun.rs
@@ -124,10 +124,27 @@ fn main() -> Result<()> {
         let sysinfo = sysinfo().context("Failed to get system information")?;
         let ram_total_mib = (sysinfo.ram_total() / (1024 * 1024)) as u32;
 
-        // By default, set the microVM RAM to be 80% of the system's RAM.
-        let ram_mib = options
-            .mem
-            .unwrap_or(MiB::from((ram_total_mib as f64 * 0.8) as u32));
+        let ram_mib = if let Some(ram_mib) = options.mem {
+            ram_mib
+        } else {
+            // By default, set the microVM RAM to be 80% of the system's RAM.
+            let guest_ram = (ram_total_mib as f64 * 0.8) as u32;
+            // Ensure we leave 3072 MiB free for the host + VRAM to avoid
+            // compromising the host's stability. This mainly applies to systems
+            // with 8GB of RAM.
+            let guest_ram = if ram_total_mib
+                .checked_sub(guest_ram)
+                .context("Failed to calculate the amount of free RAM")?
+                < 3072
+            {
+                ram_total_mib
+                    .checked_sub(3072)
+                    .context("Systems with less than 3072 MiB of RAM are not supported")?
+            } else {
+                guest_ram
+            };
+            MiB::from(guest_ram)
+        };
         // VRAM is actually the SHM window for virtio-gpu. The userspace drivers in the
         // guest are expected to be conservative and tell applications that the amount
         // of VRAM is just a percentage of the guest's RAM, so we can afford being more
--- a/crates/krun/src/bin/krun.rs
+++ b/crates/krun/src/bin/krun.rs
@@ -1,8 +1,8 @@
+use std::env;
 use std::ffi::{c_char, CString};
 use std::io::Write;
 use std::os::fd::{IntoRawFd, OwnedFd};
 use std::path::Path;
-use std::{cmp, env};
 
 use anyhow::{anyhow, Context, Result};
 use krun::cli_options::options;
@@ -12,7 +12,7 @@ use krun::launch::{launch_or_lock, LaunchResult};
 use krun::net::{connect_to_passt, start_passt};
 use krun::types::MiB;
 use krun_sys::{
-    krun_add_disk, krun_add_vsock_port, krun_create_ctx, krun_set_env, krun_set_gpu_options,
+    krun_add_disk, krun_add_vsock_port, krun_create_ctx, krun_set_env, krun_set_gpu_options2,
     krun_set_log_level, krun_set_passt_fd, krun_set_root, krun_set_vm_config, krun_set_workdir,
     krun_start_enter, VIRGLRENDERER_DRM, VIRGLRENDERER_THREAD_SYNC,
     VIRGLRENDERER_USE_ASYNC_FENCE_CB, VIRGLRENDERER_USE_EGL,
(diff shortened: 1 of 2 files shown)


alyssarosenzweig:
> Instead of trying to be clever, just configure the vRAM aperture to be 80% of the total amount of RAM in the host, independently of the amount of guest RAM configured, hoping the latter can be swapped out by the host if we run out of memory.

this seems potentially dicey, although I guess it's not really worse than the non-VM case

crates/krun/src/bin/krun.rs
```diff
@@ -120,13 +120,21 @@ fn main() -> Result<()> {
                 .or_else(|_err| get_fallback_cores())?
         };
         let num_vcpus = cpu_list.iter().fold(0, |acc, cpus| acc + cpus.len()) as u8;
+
+        let sysinfo = sysinfo().context("Failed to get system information")?;
+        let ram_total = sysinfo.ram_total() / 1024 / 1024;
```
alyssarosenzweig, line 125: / (1024 * 1024) would be much clearer.

crates/krun/src/bin/krun.rs
```diff
@@ -120,13 +120,21 @@ fn main() -> Result<()> {
...
-            let sysinfo = sysinfo().context("Failed to get system information")?;
-            let ram_total = sysinfo.ram_total() / 1024 / 1024;
             cmp::min(MiB::from((ram_total as f64 * 0.8) as u32), MiB::from(32768))
         };
+        let vram_mib = if let Some(vram_mib) = options.vram {
+            vram_mib
+        } else {
+            cmp::min(MiB::from((ram_total as f64 * 0.8) as u32), MiB::from(32768))
+        };
+
```
alyssarosenzweig, line 137:
```
let default_vram =  cmp::min(MiB::from((ram_total as f64 * 0.8) as u32), MiB::from(32768));
let vram_mib = options.vram.unwrap_or(default_vram)m;
```

alyssarosenzweig:
This is breaking Control on a 32GB machine, kcompactd gets very busy trying to get in game.

alyssarosenzweig:
> This is breaking Control on a 32GB machine, kcompactd gets very busy trying to get in game.

Dropping to 40% lets Control get in game and still exposes 12GB of VRAM.

BTW, how can the guest know the VRAM size? Honeykrisp currently assumes all RAM is VRAM, which is wrong for the VM apparently. so that's going to cause problems if it isn't already.

asahilina:
I'm confused, why does this need to be a heuristic? Isn't this just a window? It's not like VRAM is pre-allocated. Something is fishy here if we need to set a conservative value to make things work. It should be fine to set this to like, a terabyte, as long as the physical AS exists.

Also I expect there will be fragmentation problems here, so it's not strictly a VRAM limit, we should oversize it anyway to reduce the impact of fragmentation.

slp:
> I'm confused, why does this need to be a heuristic? Isn't this just a window? It's not like VRAM is pre-allocated. Something is fishy here if we need to set a conservative value to make things work. It should be fine to set this to like, a terabyte, as long as the physical AS exists. There's no reason to "limit" VRAM via krun just like the host makes no attempt to "limit" VRAM for host apps. Once you run out of physical RAM + non-VRAM swap things will start to fail either way.
> 
> Also I expect there will be fragmentation problems here, so it's not strictly a VRAM limit, we should oversize it anyway to reduce the impact of fragmentation.

Yeah, I was giving the VMM the role of restricting VRAM consumption, and that's probably not its role, as the host wouldn't do that for other GPU userspace apps running bare metal.

Now the question is who should be taking that role, since both my experience on the M1 and @alyssarosenzweig tests with Control seem to point that some applications may attempt to use as much VRAM as its available (which seems like the right thing for them to do). Should the Mesa driver somehow try to limit VRAM ram usage (actual usage, not the size of SHM window) to some value? How is this done in bare-metal to avoid running out of physical RAM to be used as actual RAM?

asahilina:
That's why I'm saying something is fishy, this isn't supposed to happen. I guess some games might try to dynamically control VRAM usage, but I'm not sure how common that is, and if anything that should be based on whatever some userspace API (GL/Vulkan) reports as "VRAM size", not anything under the hood. I'm pretty sure they wouldn't just allocate until allocations fail, that could cause everything to crash for other reasons since the game itself isn't the only thing doing VRAM allocations.

Given what I saw with Stray yesterday, where it exhausted the 8GB VRAM window before even going in-game while the game itself claimed it was using 0.2GB of VRAM, I suspect we have a leak or bloat issue somewhere...

> How is this done in bare-metal to avoid running out of physical RAM to be used as actual RAM?

It just isn't...

slp:
> > How is this done in bare-metal to avoid running out of physical RAM to be used as actual RAM?
> 
> It just isn't...

It looks like this it's actually done on macOS via the `iogpu.wired_limit_mb` sysctl tunable, which if left untouched will cap VRAM to ~65% of RAM. I don't know if the way this works is by returning the adjusted value to Metal userspace apps, or by denying allocations after reaching the limit (or both).

We can add logic in libkrun to keep track of memory allocations in the SHM region and cap it to some limit, independently of the window size, but probably userspace should also be made aware of this limit. Perhaps we could return the adjusted VRAM value as part of the DRM native context capabilities?

asahilina:
If we want to limit global VRAM size, this should be done in the host driver (in fact, probably in shared drm-shmem infrastructure), not in the guest. Though I'm wary of the whole concept because it sounds like it will strictly make things worse in most cases (if there is a global VRAM limit, then running out is likely to crash the compositor and random apps). This needs to be carefully considered before we introduce such a concept.

A *per app* VRAM limit would make more sense as a safety net, but then it should rarely be reached. That might make sense inside the VM too, though it's less important because the VM will usually be running a single game anyway (plus Steam which is a memory hog, but you don't care if the whole VM crashes if the game crashes most of the time anyway).

Part of the issue is GPU memory is not swappable right now, and that is a known limitation of the host driver. It'll be fixed at some point (that's one thing we need the non-shared BO fix for!) but we shouldn't paper over this with arbitrary VRAM limits.

But all that aside... whatever is happening with Control isn't something we should fix by reducing the window size here. We need to figure out exactly what is going on there.

slp:
From the fact that Control was working fine without this PR (and thus, with libkrun's default limit of 8GB for the SHM window), we can infer that:

- Control's engine will try to use as much VRAM as its available (and per @alyssarosenzweig's comment that's now the same as the amount of RAM in the system).
- Control's engine is clever enough to backtrack if an allocation fails.

With this in mind, I'd said the most straightforward approach would be:

1. In krun set the SHM window size to an arbitrarily large value.
2. In Mesa, adjust VRAM to be a percentage of the RAM (which would actually be guest RAM, not the actual RAM of the system).

With the default's krun behavior, which sets guest RAM to be 80% of host RAM, and with a VRAM of, let's say, 70% of the guest RAM, we would be overcommitting to the host RAM to ~140%, which is pretty reasonable having some swap configured in the system.

alyssarosenzweig:
I've adjusted things on the honeykrisp side, control seems to work again. No objection to merging this.

crates/krun/src/bin/krun.rs
slp: done, thanks

crates/krun/src/bin/krun.rs
slp: applied, thanks

teohhanhui:
Personal experience so far:

I used to be able to run Civ6 just fine (granted, that's with WineD3D)... Now it crashes after about a minute, even while just hanging around in the main menu.

PR #71, opened 2024-10-01
Prepare for renaming project to muvm

slp:
krun was already used by crun for running containers with virtualization isolated powered by libkrun. To avoid confusing, let's rename this one to muvm.


teohhanhui:
What does it stand for? 😛

alyssarosenzweig:
> What does it stand for? 😛

microVM https://en.wikipedia.org/wiki/Micro-

alyssarosenzweig (approved):

slp:
Let's also bump the version to avoid confusion.

PR #72, opened 2024-10-03
Basic x112virtgpu support

asahilina:
Replace the trivial X11 socket passthrough with x112virtgpu, if it is installed and DAX is enabled. This lives alongside the existing sommelier solution, so we can test both options for now.

### DIFF
--- a/crates/muvm/src/cli_options.rs
+++ b/crates/muvm/src/cli_options.rs
@@ -15,6 +15,7 @@ pub struct Options {
     pub passt_socket: Option<PathBuf>,
     pub server_port: u32,
     pub fex_images: Vec<String>,
+    pub direct_x11: bool,
     pub command: PathBuf,
     pub command_args: Vec<String>,
 }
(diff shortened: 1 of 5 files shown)


alyssarosenzweig (approved):

slp (changes requested):
Small change requested. Also, please add a SoB to the second commit.

asahilina:
Removed the DAX check and snuck in some Clippy fixes in related code.

This now soft-depends on #96. They can be merged in any order, but x112virtgpu will only work as intended if #96 is also merged.

alyssarosenzweig (approved):
les go

asahilina:
Added another commit to make this mutually exclusive with sommelier, and enabled with the `--direct-x11` option. This makes it less confusing to test, and there's no real reason to aim to support sommelier+this simultaneously long term.

crates/muvm/src/cli_options.rs
```diff
@@ -98,6 +99,11 @@ pub fn options() -> OptionParser<Options> {
         .argument("SERVER_PORT")
         .fallback(3334)
         .display_fallback();
+    let direct_x11 = long("direct-x11")
+        .short('x')
+        .help("Use direct X11 forwarding instead of sommelier + XWayland")
+        .flag(true, false);
```
teohhanhui, line 105:
```suggestion
        .switch();
```

https://docs.rs/bpaf/latest/bpaf/parsers/struct.NamedArg.html#method.switch

asahilina: Fixed, thanks!

teohhanhui (approved):

slp (approved):

PR #74, opened 2024-10-07
Hotfix for missing /etc/hostname on new gnome installs

alyssarosenzweig:

PR #75, opened 2024-10-08
guest/net: Handle missing /etc/hostname gracefully

jannau:
The error handled in #74 was muvm-guest's network code. So fixing it there directly. Not sure if any of the actual binaries running in the guest actually care about `/etc/hostname`.


teohhanhui (approved):

alyssarosenzweig (approved):

slp:
Should we revert #74 or it's actually safer keeping it around?

PR #86, opened 2024-10-13
move_mount if /etc/resolv.conf is symlink

### LINKED ISSUE #85: Doesn't launch, crashes.
Hi, first of all wanted to thank you for all your contributions to the mac ecosystem.

So, I don't really don't know what logs to provide(but please do tell), but it errors out with this:
```

vsevolodlavrusha@fedora:~$ muvm -- sh
Error: Failed to mount filesystems, bailing out

Caused by:
    0: Failed to move_mount tmp to etc
    1: No such file or directory (os error 2)

```

IsaacMarovitz:
Should resolve #85 (same issue occurs on my system).

### DIFF
--- a/crates/muvm/src/guest/mount.rs
+++ b/crates/muvm/src/guest/mount.rs
@@ -122,7 +122,7 @@ pub fn place_etc(file: &str, contents: Option<&str>) -> Result<()> {
         "",
         CWD,
         etc,
-        MoveMountFlags::MOVE_MOUNT_F_EMPTY_PATH,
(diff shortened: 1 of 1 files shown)


slp:
In my system, `/etc/resolv.conf` is a symlink but I don't have an issue here without this PR. How does the issue manifest in your system?

teohhanhui:
Looking at the docs: https://github.com/sunfishcode/linux-mount-api-documentation/blob/main/move_mount.md

I don't understand why it matters whether `/etc/resolv.conf` is a symlink or not. The `move_mount` call is moving the mount from `/tmp/resolv.conf` to `/etc/resolv.conf`, right? What does it mean to follow the destination symlink? Is it applying to `/etc` instead?

PR #87, opened 2024-10-15
muvm-guest: Support root and user init scripts

asahilina:
Introduce two scripts that can be placed in ~/.local/share/muvm

- init.sh: Executed as root, before dropping privileges
- user.sh: Executed as the user, after dropping privileges

If the files are also present in lower-priority XDG dirs (systemwide), those are executed first.

This makes it possible to add custom VM configuration and setup or debugging scripts without having to hack on the muvm code.

### DIFF
--- a/crates/muvm/src/guest/bin/muvm-guest.rs
+++ b/crates/muvm/src/guest/bin/muvm-guest.rs
@@ -1,4 +1,5 @@
 use std::cmp;
+use std::env;
 use std::os::unix::process::CommandExt as _;
 use std::process::Command;
 
@@ -13,6 +14,7 @@ use muvm::guest::sommelier::exec_sommelier;
 use muvm::guest::user::setup_user;
 use muvm::guest::x11::setup_x11_forwarding;
 use muvm::utils::env::find_in_path;
+use nix::unistd::User;
 use rustix::process::{getrlimit, setrlimit, Resource};
 
 fn main() -> Result<()> {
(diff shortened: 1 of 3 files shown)


crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -40,6 +42,18 @@ fn main() -> Result<()> {
...
 
+    let user = User::from_uid(options.uid)?.unwrap();
+
+    // Hack to fetch the XDG directories for our target user, before we
+    // change uid.
+    // SAFETY: Safe if and only if `muvm-guest` program is not multithreaded.
+    // See https://doc.rust-lang.org/std/env/fn.set_var.html#safety
+    env::set_var("HOME", user.dir);
+    let xdg_dirs = xdg::BaseDirectories::with_prefix("muvm")?;
+    // SAFETY: Safe if and only if `muvm-guest` program is not multithreaded.
```
alyssarosenzweig, line 53: cc @slp for #64 interaction

slp:
I think I understand the motivation, but I'm afraid this may open the floods to users implementing all shorts of hacks behind `muvm-guest`'s back. If it's intended just for developing purposes, the role of `user.sh` could be achieved by calling it as the first program in the guest, and for the role of `root.sh`, #89 will allow launching programs as root in the guest (and, combined with #48, it'll also be possible to have a root shell).

Beyond developing purposes, I think the configuration to the VM should be applied by `muvm-guest` itself.

asahilina:
Closing in favor of #89

PR #88, opened 2024-10-15
muvm-guest: Reopen the system tty (hvc0)

asahilina:
This fixes the classic:

-bash: cannot set terminal process group (-1): Inappropriate ioctl for device
-bash: no job control in this shell

I'm not entirely sure why this doesn't happen with our main shell, but it happens with things like sudo. (There are a bunch of other things needed to make sudo work, of course)

### DIFF
--- a/crates/muvm/src/guest/bin/muvm-guest.rs
+++ b/crates/muvm/src/guest/bin/muvm-guest.rs
@@ -1,4 +1,5 @@
 use std::cmp;
+use std::os::fd::AsFd;
 use std::os::unix::process::CommandExt as _;
 use std::process::Command;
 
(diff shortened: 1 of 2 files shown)


alyssarosenzweig (approved):

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -13,6 +13,7 @@ use muvm::guest::sommelier::exec_sommelier;
 use muvm::guest::user::setup_user;
 use muvm::guest::x11::setup_x11_forwarding;
 use muvm::utils::env::find_in_path;
+use rustix::fd::AsFd;
```
teohhanhui, line 16: This is just a [re-export](https://docs.rs/rustix/latest/rustix/fd/trait.AsFd.html) of [`std::os::fd::AsFd`](https://doc.rust-lang.org/std/os/fd/trait.AsFd.html)

slp (changes requested):
LGTM once the issue pointed by @teohhanhui is addressed.

asahilina:
Fixed.

slp (approved):

PR #89, opened 2024-10-15
Protect muvm-server, spawn one as root and monitor available host memory

slp:
The goal of this PR is to enable muvm to request the guest to adjust its file system cache goals accordingly to the amount of available memory on the host.

The first two commits just set the infrastructure, first by adding some minimal protection to muvm-server, and then spawning a second instance of it as root (we need this to operate in "/proc/sys/vm", but surely will come handy for many other things).

The latter implements the monitor and the ability to request the guest to adjust watermark_scale_factor as needed, or even to drop its pagecache if the situation is critical.

### DIFF
--- a/crates/muvm/src/guest/bin/muvm-guest.rs
+++ b/crates/muvm/src/guest/bin/muvm-guest.rs
@@ -62,6 +62,12 @@ fn main() -> Result<()> {
             .spawn()?;
     }
 
+    // Before switching to the user, start another instance of muvm-server to serve
+    // launch requests as root.
+    Command::new("muvm-server")
+        .spawn()
+        .context("Failed to execute `muvm-server` as child process")?;
+
     let run_path = match setup_user(options.username, options.uid, options.gid) {
         Ok(p) => p,
         Err(err) => return Err(err).context("Failed to set up user, bailing out"),
--- a/crates/muvm/src/launch.rs
+++ b/crates/muvm/src/launch.rs
@@ -148,7 +148,7 @@ fn lock_file() -> Result<(Option<File>, Uuid)> {
     Ok((Some(lock_file), cookie))
 }
 
-fn request_launch(
+pub fn request_launch(
     server_port: u32,
     cookie: Uuid,
     command: PathBuf,
--- a/crates/muvm/src/launch.rs
+++ b/crates/muvm/src/launch.rs
@@ -2,14 +2,14 @@ use std::collections::HashMap;
 use std::env;
 use std::error::Error;
 use std::fmt::{Display, Formatter};
-use std::fs::File;
+use std::fs::{self, File};
 use std::io::{BufRead, BufReader, Read, Write};
 use std::net::TcpStream;
 use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Context, Result};
 use rustix::fs::{flock, FlockOperation};
-use rustix::path::Arg;
+use uuid::Uuid;
 
 use crate::env::prepare_env_vars;
 use crate::utils::launch::Launch;
@@ -17,6 +17,7 @@ use crate::utils::launch::Launch;
 pub enum LaunchResult {
     LaunchRequested,
     LockAcquired {
+        cookie: Uuid,
         lock_file: File,
         command: PathBuf,
         command_args: Vec<String>,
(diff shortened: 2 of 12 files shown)


crates/muvm/src/monitor.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl From<GuestPressure> for u32 {
+    fn from(pressure: GuestPressure) -> u32 {
+        match pressure {
+            GuestPressure::None => 10,
+            GuestPressure::Low => 1000,
+            GuestPressure::Medium => 2000,
+            GuestPressure::High => 3000,
+            GuestPressure::Critical => 3000,
```
teohhanhui, line 29: Same values intentional?

crates/muvm/src/monitor.rs
```diff
@@ -0,0 +1,94 @@
...
+    let wsf: u32 = pressure.into();
+    debug!("setting watermark_scale_factor to {wsf}");
+
+    let command = PathBuf::from("/sbin/sysctl");
+    let command_args = vec![format!("vm.watermark_scale_factor={}", wsf)];
+    let env = HashMap::new();
+    request_launch(server_port, cookie, command, command_args, env)
+}
+
+fn work(server_port: u32, cookie: Uuid) {
```
teohhanhui, line 57: Not a verb in this case (it'd have to be "do work")... What about `run`?

crates/muvm/src/net.rs
```diff
@@ -41,6 +41,8 @@ pub fn start_passt(server_port: u32) -> Result<UnixStream> {
     let child = Command::new("passt")
         .args(["-q", "-f", "-t"])
         .arg(format!("{server_port}:{server_port}"))
+        .arg("-t")
+        .arg(format!("{root_server_port}:{root_server_port}"))
```
teohhanhui, line 45: Should be combine with the above `-t`:

From the `passt` manpage:

> A comma-separated list of ports

crates/muvm/src/server/bin/muvm-server.rs
```diff
@@ -1,31 +1,58 @@
...
+
+    let rt = tokio::runtime::Runtime::new().unwrap();
+    rt.block_on(async { tokio_main(cookie).await })
+}
+
+async fn tokio_main(cookie: String) -> Result<()> {
     env_logger::init();
 
-    let options = options().run();
+    let cookie = Uuid::from_str(&cookie).context("Couldn't parse cookie as UUID v7")?;
```
teohhanhui, line 29: [`Uuid::from_str`](https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.from_str) just calls [`Uuid::parse_str`](https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.parse_str), and the docs also mentions to use [`Uuid::try_parse`](https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.try_parse) unless we need detailed error information, which I don't think we do in this case?

crates/muvm/src/monitor.rs
slp: Makes sense, I've just added it.

crates/muvm/src/monitor.rs
slp: `run` sounds good to me, changed.

crates/muvm/src/net.rs
slp: Done.

crates/muvm/src/server/bin/muvm-server.rs
slp: You're right, `try_parse` it's better. Changed.

alyssarosenzweig (approved):
lgtm, but probably should get review/testing from someone more rust-brained

teohhanhui:
Will take another look... 🙇

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -50,6 +50,12 @@ fn main() -> Result<()> {
             .spawn()?;
     }
 
+    // Before switching to the user, start another instance of muvm-server to serve
+    // launch requests as root.
+    if let Some(muvm_server_path) = find_in_path("muvm-server")? {
+        Command::new(muvm_server_path).spawn()?;
+    }
```
teohhanhui, line 57: Are there situations where `muvm-server` is expected to be missing? If not this could just be:

```suggestion
    Command::new("muvm-server").spawn().context("Failed to find `muvm-server`")?;
```

crates/muvm/src/launch.rs
```diff
@@ -59,53 +60,69 @@ pub fn launch_or_lock(
...
+        .context("Failed to read XDG_RUNTIME_DIR environment variable")?;
+    let lock_path = Path::new(&run_path).join("muvm.lock");
+    let mut lock_file = File::options()
+        .write(false)
+        .read(true)
+        .open(lock_path)
+        .context("Failed to open lock file")?;
+
+    let mut data: Vec<u8> = Vec::with_capacity(16);
+    lock_file.read_to_end(&mut data)?;
```
teohhanhui, line 120:
```suggestion
    let data: Vec<u8> = fs::read(lock_path).context("Failed to read lock file")?;
```

https://doc.rust-lang.org/std/fs/fn.read.html

teohhanhui: As a sanity check, we could check the `data.len()`?

crates/muvm/src/server/worker.rs
```diff
@@ -158,15 +171,42 @@ async fn read_request(stream: &mut BufStream<TcpStream>) -> Result<Launch> {
...
     debug!(command:?, command_args:?, env:?; "received launch request");
+    if cookie != server_cookie {
+        debug!("invalid cookie in launch request");
+        let msg = "Invalid cookie";
+        stream.write_all(msg.as_bytes()).await.ok();
+        stream.flush().await.ok();
+        return Err(anyhow!(msg));
+    }
+
+    if command.to_string_lossy().contains("/muvmdropcaches") {
```
teohhanhui, line 195:
```suggestion
    if command == Path::new("/muvmdropcaches") {
```

crates/muvm/src/guest/bin/muvm-guest.rs
slp: Applied, thanks.

crates/muvm/src/launch.rs
slp: Nice suggestions, thanks!

crates/muvm/src/server/worker.rs
slp: Done.

teohhanhui (approved):

PR #90, opened 2024-10-16
Do not clobber FEX config if no images were mounted.

WhatAmISupposedToPutHere:
On Asahi Gentoo the FEX rootfs is a part of the host filesystem, and as such the muvm image mount mechanism is not used. Skip overmounting the FEX directory in that case.


teohhanhui (approved):
LGTM

alyssarosenzweig (approved):

PR #92, opened 2024-10-20
ci: Upgrade checkout to v4

Margen67:


teohhanhui (approved):
Checked the changelog. No relevant breaking changes for us.

PR #97, opened 2024-10-31
Move in x112virtgpu as muvm-x11bridge

asahilina:
Per discussion with @WhatAmISupposedToPutHere and @slp, let's move x112virtgpu development here since muvm is the only consumer.


WhatAmISupposedToPutHere:
Acked-by: Sasha Finkelstein <fnkl.kernel@gmail.com>

crates/x11bridge/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "muvm-x11bridge"
```
teohhanhui, line 2: Shall we rename the directory to `muvm-x11bridge` too to match the package name?

crates/x11bridge/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.0.1"
+edition = "2021"
+
+[dependencies]
+nix = { version = "0.29", features = ["event", "socket", "ioctl", "mman", "socket", "uio", "fs", "ptrace", "signal", "user"] }
+anyhow = "1"
```
teohhanhui, line 8:
```suggestion
nix = { version = "0.29.0", default-features = false, features = ["event", "fs", "ioctl", "mman", "ptrace", "signal", "socket", "uio", "user"] }
anyhow = { version = "1.0.91", default-features = false, features = ["std"] }
```

crates/x11bridge/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+bpaf = { version = "0.9.11", default-features = false, features = [] }
+
+[[bin]]
+name = "muvm-x11bridge"
+path = "src/bin/muvm-x11bridge.rs"
```
teohhanhui, line 13: This is already the auto-detected default...

crates/x11bridge/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "muvm-x11bridge"
+version = "0.0.1"
+edition = "2021"
+
+[dependencies]
+nix = { version = "0.29", features = ["event", "socket", "ioctl", "mman", "socket", "uio", "fs", "ptrace", "signal", "user"] }
+anyhow = "1"
+bpaf = { version = "0.9.11", default-features = false, features = [] }
```
teohhanhui, line 9:
```suggestion
anyhow = { version = "1.0.91", default-features = false, features = ["std"] }
bpaf = { version = "0.9.11", default-features = false, features = [] }
nix = { version = "0.29.0", default-features = false, features = ["event", "fs", "ioctl", "mman", "ptrace", "signal", "socket", "uio", "user"] }
```

crates/x11bridge/Cargo.toml
asahilina: Done!

crates/x11bridge/Cargo.toml
asahilina: Thanks, applied!

alyssarosenzweig (approved):

crates/muvm-x11bridge/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "muvm-x11bridge"
+version = "0.0.1"
+edition = "2021"
```
teohhanhui, line 4:
```suggestion
name = "muvm-x11bridge"
version = "0.0.1"
authors = [...]
edition = "2021"
rust-version = "1.77.0"
license = "..."
```

Some missing fields that should probably be filled in?

crates/muvm/Cargo.toml
```diff
@@ -15,7 +15,7 @@ byteorder = { version = "1.5.0", default-features = false, features = ["std"] }
 env_logger = { version = "0.11.3", default-features = false, features = ["auto-color", "humantime", "unstable-kv"] }
 krun-sys = { path = "../krun-sys", version = "1.9.1", default-features = false, features = [] }
 log = { version = "0.4.21", default-features = false, features = ["kv"] }
-nix = { version = "0.28.0", default-features = false, features = ["user"] }
+nix = { version = "0.29.0", default-features = false, features = ["user"] }
```
teohhanhui, line 18: Should we move the shared dependencies versions back to `[workspace.dependencies]` in the root `Cargo.toml`?

/cc @slp :laughing:

crates/muvm-x11bridge/Cargo.toml
slp: I agree. Also, please change the version to `0.1.0` so it's easier to follow SemVer.

slp:
Do we want to publish `muvm-x11bridge` independently? If not (and I'd strongly suggest against it, given the past experience with `muvm-guest` and `muvm-server`), I'd suggest putting it under `crates/muvm/src/x11bridge` and to follow the same organization as `muvm-guest` and `muvm-server`. This way `muvm`, composed by the `muvm`, `muvm-guest`, `muvm-server` and `muvm-x11bridge` binaries, will be published as a single crate to crates.io and as a single package in Fedora.

asahilina:
FWIW I also prototyped optionally dropping in sommelier for pure Wayland forwarding, but it's *quite* broken. It seems `--parent` mode does not work at all with `--virtgpu-channel` (it tries to open `/dev/wl0` and fails). Per-app mode can be tested by just doing stuff like `/usr/bin/sommelier --virtgpu-channel --log-level=3 firefox` and seems to mostly work, but not all apps support the fd-based Wayland connection that requires, and it needs a per-app launch and doesn't support apps launching other apps, so it's basically a non-starter.

So I suggest we just leave Wayland out of muvm for now until we have a sommelier alternative. People who want to try the fd-based mode can just launch it explicitly with an app.

slp (approved):
LGTM

PR #104, opened 2024-11-05
Improve VRAM management in the guest

slp:
By default, HK sets the heap size to be half the size of the *guest*
    memory. Since commit 167744dc it's also possible to override the heap
    size by setting the HK_SYSMEM environment variable.

    Previously, the vram command line option was acting on the virtio-gpu
    SHM window. This didn't help in most cases because userspace doesn't
    really care nor is aware of it.

    Here, we set the SHM window of virtio-gpu to always be as large as the
    host's RAM, not because we expect VRAM to consume 100% of the RAM (which
    is obviously impossible) but to account for region fragmentation. Then,
    we use HK_SYSMEM to tell userspace the desired heap size (the closest
    thing to VRAM we can aim for), which will be set to either the value
    passed by the user throught the "vram" argument or half the size of the
    *host* RAM.

### DIFF
--- a/crates/muvm/src/bin/muvm.rs
+++ b/crates/muvm/src/bin/muvm.rs
@@ -149,12 +149,22 @@ fn main() -> Result<()> {
             };
             MiB::from(guest_ram)
         };
-        // VRAM is actually the SHM window for virtio-gpu. The userspace drivers in the
-        // guest are expected to be conservative and tell applications that the amount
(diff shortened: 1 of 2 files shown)


alyssarosenzweig (approved):
I don't love the HK backdoor but until we have better kernel level stuff this seems reasonable +1

PR #105, opened 2024-11-13
env: Pass through locale & IM stuff

asahilina:
Leaving the default C locale can break stuff that expect UTF-8, so pass through all the locale env vars and set a default C.UTF-8 locale if nothing is set.

For input methods, it turns out that the old legacy XIM does work with our X11 passthrough solution (direct GTK/QT plugin methods will not, since they use side-channels to communicate with the IM server). So, force xim usage for GTK2/3 and QT4/5, which actually works with Steam (!) for CJK input if the system is configured properly.

The future is Wayland protocols, which will work with GTK4 and QT6 once we have Wayland passthrough. Hopefully that means we never have to support input methods with the shared library mechanisms, which means our rootfs will never have to include IM-specific libraries and the input method user choice will happen entirely outside of muvm.


crates/muvm/src/env.rs
```diff
@@ -10,12 +10,27 @@ use log::debug;
...
 /// Automatically pass these environment variables to the microVM, if they are
 /// set.
-const WELL_KNOWN_ENV_VARS: [&str; 5] = [
+const WELL_KNOWN_ENV_VARS: [&str; 20] = [
     "LD_LIBRARY_PATH",
     "LIBGL_DRIVERS_PATH",
     "MESA_LOADER_DRIVER_OVERRIDE", // needed for asahi
     "PATH",                        // needed by `muvm-guest` program
     "RUST_LOG",
+    "LANG",
```
teohhanhui, line 19: Sort alphabetically?

teohhanhui (approved):

slp:
LGTM, but please address @teohhanhui suggestion.

asahilina:
Sorted.

asahilina:
Sure, maybe we should merge just that commit for now? @WhatAmISupposedToPutHere what do you think? I can rebase this PR on just that one commit and then add the xim / extra locale handling on top.

WhatAmISupposedToPutHere:
Feel free to use that one. Sorry for getting distracted with shinier stuff and leaving that branch as is, it is somewhere on the yak stack and i will get back to it one day.

teohhanhui:
> just pass almost everything through

That kinda makes me nervous. Haha... The chances of some software misbehaving and messing up something in the user's home directory seems non-negligible?

WhatAmISupposedToPutHere:
I do not think it is that likely, most /home contents are arch-independent, and we already allow writing there.

alyssarosenzweig:
> > just pass almost everything through
> 
> That kinda makes me nervous. Haha... The chances of some software misbehaving and messing up something in the user's home directory seems non-negligible?

TBH I don't see how the situation is different from today. Env vars aren't particularly special..?

asahilina:
Turns out I can't cherry-pick that commit cleanly since it's passing through all env vars to the process handler, it doesn't do it to the VM itself... and that whole idea depends on earlier commits.

Maybe we can just merge this one for now?

slp:
> Turns out I can't cherry-pick that commit cleanly since it's passing through all env vars to the process handler, it doesn't do it to the VM itself... and that whole idea depends on earlier commits.
> 
> Maybe we can just merge this one for now?

Ack, let's merge this one and work later on #54

PR #106, opened 2024-11-15
guest: Do not crash if no GPU or sound devices

hoshinolina:
This can happen if libkrun was built without those flags.


sbrivio-rh:
Tested, thanks, it works for me (libkrun built without `SND=1`, without `GPU=1`).

teohhanhui (approved):

PR #107, opened 2024-11-18
hidpipe: merge hidpipe project into muvm

slp:
Merge hidpipe project into muvm, with hidpipe-server operating as a thread inside the main muvm binary, and hidpipe-client getting renamed to muvm-hidpipe (and it's still autostarted by muvm-guest).

This simplifies the lifecycle management for hidpipe, ensuring it's tied to the microVM lifetime.

Co-developed-by: Sasha Finkelstein <fnkl.kernel@gmail.com>

### DIFF
--- a/crates/muvm/Cargo.toml
+++ b/crates/muvm/Cargo.toml
@@ -13,6 +13,8 @@ anyhow = { version = "1.0.82", default-features = false, features = ["std"] }
 bpaf = { version = "0.9.11", default-features = false, features = [] }
 byteorder = { version = "1.5.0", default-features = false, features = ["std"] }
 env_logger = { version = "0.11.3", default-features = false, features = ["auto-color", "humantime", "unstable-kv"] }
+input-linux = { version = "0.7.0", default-features = false, features = [] }
+input-linux-sys = { version = "0.9.0", default-features = false, features = [] }
 krun-sys = { path = "../krun-sys", version = "1.9.1", default-features = false, features = [] }
 log = { version = "0.4.21", default-features = false, features = ["kv"] }
 nix = { version = "0.29.0", default-features = false, features = ["user"] }
@@ -23,6 +25,7 @@ serde_json = { version = "1.0.117", default-features = false, features = ["std"]
 tempfile = { version = "3.10.1", default-features = false, features = [] }
 tokio = { version = "1.38.0", default-features = false, features = ["io-util", "macros", "net", "process", "rt-multi-thread", "sync"] }
 tokio-stream = { version = "0.1.15", default-features = false, features = ["net", "sync"] }
+udev = { version = "0.9.0", default-features = false, features = [] }
 uuid = { version = "1.10.0", default-features = false, features = ["serde", "std", "v7"] }
 
 [features]
@@ -36,6 +39,10 @@ path = "src/bin/muvm.rs"
 [[bin]]
 name = "muvm-guest"
 path = "src/guest/bin/muvm-guest.rs"
+
+[[bin]]
+name = "muvm-hidpipe"
+path = "src/hidpipe/bin/muvm-hidpipe.rs"
 
 [[bin]]
 name = "muvm-server"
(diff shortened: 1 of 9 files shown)


slp:
@WhatAmISupposedToPutHere PTAL

WhatAmISupposedToPutHere (approved):

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -57,7 +57,7 @@ fn main() -> Result<()> {
 
     configure_network()?;
 
-    if let Some(hidpipe_client_path) = find_in_path("hidpipe-client")? {
+    if let Some(hidpipe_client_path) = find_in_path("muvm-hidpipe")? {
```
alyssarosenzweig, line 60: Now thta we're bundling, we should drop this check.

crates/muvm/src/bin/muvm.rs
```diff
@@ -297,13 +298,14 @@ fn main() -> Result<()> {
             }
         }
         let hidpipe_path = Path::new(&run_path).join("hidpipe");
-        if hidpipe_path.exists() {
+        if spawn_hidpipe_server(hidpipe_path.clone()).is_ok() {
```
alyssarosenzweig, line 301: Should this still be allowed to fail?

teohhanhui:
Could this be put behind a `hidpipe` feature gate?

crates/muvm/src/bin/muvm.rs
slp: Sounds reasonable to me.

slp:
> Could this be put behind a `hidpipe` feature gate?

It's possible, but if no distros plan to build `muvm` without `hidpipe` seems like an unneeded complexity to me...

crates/muvm/src/guest/bin/muvm-guest.rs
alyssarosenzweig: :-/ Seems like a packaging issue if we can't find it though. would be inclined to make this a hard req instead of getting weird "gamepads don't work on Ubuntu Asahi help!!" bugs when it could've been caught earlier on

teohhanhui: The flip side is that not everyone uses gamepads / controllers, and we have already seen in some issues here that there are people who want to run muvm headless. So I think it actually makes sense to have this feature gated.

asahilina: Is it really necessary to feature gate it at the source level though? Distros probably aren't going to split-package this, it's  kind of pointless... for people who don't use gamepads/controllers, it should just do nothing at runtime.

slp: I agree with @asahilina here. Gating stuff has a maintenance cost, so there must be a benefit that justifies it, and I don't think enabling distros to disable hidpipe does it. In fact, I think x11bridge should be ungated as soon as it's made the default (which, based on my experience testing it, should be done fairly soon).

I'm going to change the `find_in_path` with a direct `Command::new()`, as we're doing with `muvm-server` and @alyssarosenzweig suggested.

crates/muvm/src/bin/muvm.rs
slp: For consistency with `muvm-guest.rs` treatment of `muvm-hidpipe`, changed to fail if can't be spawned.

slp:
@alyssarosenzweig PTAAL.

crates/muvm/Cargo.toml
```diff
@@ -45,3 +48,7 @@ path = "src/server/bin/muvm-server.rs"
 name = "muvm-x11bridge"
 path = "src/x11bridge/bin/muvm-x11bridge.rs"
 required-features = ["x11bridge"]
+
+[[bin]]
+name = "muvm-hidpipe"
+path = "src/hidpipe/muvm-hidpipe.rs"
```
teohhanhui, line 54:
```suggestion
path = "src/hidpipe/bin/muvm-hidpipe.rs"
```

For consistency?

teohhanhui:
Temporarily uncommenting the commented lines in `rustfmt.toml` and running `cargo +nightly fmt` should fix up some of the CS...

alyssarosenzweig:
Ack from me, although probably want to address Teoh's comments

slp:
Merging after applying @teohhanhui 's suggestions.

PR #111, opened 2024-11-25
guest/net: New implementation of network setup with SLAAC and own DHC…

sbrivio-rh:
…P client

The existing implementation has a couple of issues:

- it doesn't support IPv6 or SLAAC

- it relies on either dhclient(8) or dhcpcd(8), which need a significant amount of time to configure the network as they are rather generic DHCP clients

- on top of this, dhcpcd, by default, unless --noarp is given, will spend five seconds ARP-probing the address it just received before configuring it

Replace the IPv4 part with a minimalistic, 73-line DHCP client that just does what we need, using option 80 (Rapid Commit) to speed up the whole exchange.

Add IPv6 support (including IPv4-only, and IPv6-only modes) relying on the kernel to perform SLAAC. Safely avoid DAD (we're the only node on the link) by disabling router solicitations, starting SLAAC, and re-enabling them once addresses are configured.

Instead of merely triggering the network setup and proceeding, wait until everything is configured, so that connectivity is guaranteed to be ready before any further process runs in the guest, say:

  $ ./target/debug/muvm -- ping -c1 2a01:4f8:222:904::2
  PING 2a01:4f8:222:904::2 (2a01:4f8:222:904::2) 56 data bytes
  64 bytes from 2a01:4f8:222:904::2: icmp_seq=1 ttl=255 time=0.256 ms

  --- 2a01:4f8:222:904::2 ping statistics ---
  1 packets transmitted, 1 received, 0% packet loss, time 0ms
  rtt min/avg/max/mdev = 0.256/0.256/0.256/0.000 ms

The whole procedure now takes approximately 1.5 to 2 ms (for both IPv4 and IPv6), with the DHCP exchange and configuration taking somewhere around 300-500 µs out of that, instead of hundreds of milliseconds to seconds.

Matching support in passt for option 80 (RFC 4039) and for the DHCP "broadcast" flag (RFC 2131) needs this series:

  https://archives.passt.top/passt-dev/20241125152812.369553-1-sbrivio@redhat.com/

[I'll update this commit message once we have an upstream release with it]

### DIFF
--- a/crates/muvm/Cargo.toml
+++ b/crates/muvm/Cargo.toml
@@ -3,7 +3,7 @@ name = "muvm"
 version = "0.3.1"
 authors = ["Sergio Lopez <slp@redhat.com>", "Teoh Han Hui <teohhanhui@gmail.com>", "Sasha Finkelstein <fnkl.kernel@gmail.com>", "Asahi Lina <lina@asahilina.net>"]
 edition = "2021"
-rust-version = "1.77.0"
+rust-version = "1.80.0"
 description = "Run programs from your system in a microVM"
 repository = "https://github.com/AsahiLinux/muvm"
 license = "MIT"
@@ -12,12 +12,14 @@ license = "MIT"
 anyhow = { version = "1.0.82", default-features = false, features = ["std"] }
 bpaf = { version = "0.9.11", default-features = false, features = [] }
 byteorder = { version = "1.5.0", default-features = false, features = ["std"] }
+const-str = { version = "0.6.2", default-features = false, features = [] }
 env_logger = { version = "0.11.3", default-features = false, features = ["auto-color", "humantime", "unstable-kv"] }
 input-linux = { version = "0.7.0", default-features = false, features = [] }
 input-linux-sys = { version = "0.9.0", default-features = false, features = [] }
 krun-sys = { path = "../krun-sys", version = "1.9.1", default-features = false, features = [] }
 log = { version = "0.4.21", default-features = false, features = ["kv"] }
 nix = { version = "0.29.0", default-features = false, features = ["event", "fs", "ioctl", "mman", "ptrace", "signal", "socket", "uio", "user"] }
+neli = { version = "0.7.0-rc3", default-features = false, features = ["sync"] }
 procfs = { version = "0.17.0", default-features = false, features = [] }
 rustix = { version = "0.38.34", default-features = false, features = ["fs", "mount", "process", "pty", "std", "stdio", "system", "termios", "use-libc-auxv"] }
 serde = { version = "1.0.203", default-features = false, features = ["derive"] }
--- a/crates/muvm/src/guest/net.rs
+++ b/crates/muvm/src/guest/net.rs
@@ -1,14 +1,266 @@
 use std::fs;
 use std::io::Write;
-use std::os::unix::process::ExitStatusExt as _;
-use std::process::Command;
+use std::net::{Ipv4Addr, UdpSocket};
+use std::time::Duration;
 
-use anyhow::{anyhow, Context, Result};
-use log::debug;
+use anyhow::{Context, Result};
+use neli::consts::nl::NlmF;
+use neli::consts::rtnl::{
+    Arphrd, Ifa, IfaF, Iff, Ifla, RtAddrFamily, RtScope, RtTable, Rta, Rtm, RtmF, Rtn, Rtprot,
+};
+use neli::consts::socket::NlFamily;
+use neli::nl::{NlPayload, Nlmsghdr};
+use neli::router::synchronous::{NlRouter, NlRouterReceiverHandle};
+use neli::rtnl::{
+    Ifaddrmsg, IfaddrmsgBuilder, Ifinfomsg, IfinfomsgBuilder, RtattrBuilder, Rtmsg, RtmsgBuilder,
+};
+use neli::types::RtBuffer;
+use neli::utils::Groups;
 use rustix::system::sethostname;
 
-use crate::utils::env::find_in_path;
-use crate::utils::fs::find_executable;
+/// Set interface flags for eth0 (interface index 2) with a given mask
+fn flags_eth0(rtnl: &NlRouter, mask: Iff, set: Iff) -> Result<()> {
+    let ifinfomsg = IfinfomsgBuilder::default()
+        .ifi_family(RtAddrFamily::Unspecified)
+        .ifi_type(Arphrd::Ether)
+        .ifi_index(2)
+        .ifi_change(mask)
+        .ifi_flags(set)
+        .build()?;
+
+    let _: NlRouterReceiverHandle<Rtm, Ifinfomsg> =
+        rtnl.send(Rtm::Newlink, NlmF::REQUEST, NlPayload::Payload(ifinfomsg))?;
+
+    Ok(())
+}
+
+/// Set MTU for eth0 (interface index 2)
+fn mtu_eth0(rtnl: &NlRouter, mtu: u32) -> Result<()> {
+    let ifinfomsg = IfinfomsgBuilder::default()
+        .ifi_family(RtAddrFamily::Unspecified)
+        .ifi_type(Arphrd::Ether)
+        .ifi_index(2)
+        .rtattrs(RtBuffer::from_iter([RtattrBuilder::default()
+            .rta_type(Ifla::Mtu)
+            .rta_payload(mtu)
+            .build()?]))
+        .build()?;
+
+    let _: NlRouterReceiverHandle<Rtm, Ifinfomsg> =
+        rtnl.send(Rtm::Newlink, NlmF::REQUEST, NlPayload::Payload(ifinfomsg))?;
+
+    Ok(())
+}
+
+/// Add or delete IPv4 routes for eth0 (interface index 2)
+fn route4_eth0(rtnl: &NlRouter, what: Rtm, gw: Ipv4Addr) -> Result<()> {
+    let rtmsg = RtmsgBuilder::default()
+        .rtm_family(RtAddrFamily::Inet)
+        .rtm_dst_len(0)
+        .rtm_src_len(0)
+        .rtm_tos(0)
+        .rtm_table(RtTable::Main)
+        .rtm_protocol(Rtprot::Boot)
+        .rtm_scope(RtScope::Universe)
+        .rtm_type(Rtn::Unicast)
+        .rtm_flags(RtmF::empty())
+        .rtattrs(RtBuffer::from_iter([
+            RtattrBuilder::default()
+                .rta_type(Rta::Oif)
+                .rta_payload(2)
+                .build()?,
+            RtattrBuilder::default()
+                .rta_type(Rta::Dst)
+                .rta_payload(Ipv4Addr::UNSPECIFIED.octets().to_vec())
+                .build()?,
+            RtattrBuilder::default()
+                .rta_type(Rta::Gateway)
+                .rta_payload(gw.octets().to_vec())
+                .build()?,
+        ]))
+        .build()?;
+
+    let _: NlRouterReceiverHandle<Rtm, Rtmsg> = rtnl.send(
+        what,
+        NlmF::CREATE | NlmF::REQUEST,
+        NlPayload::Payload(rtmsg),
+    )?;
+
+    Ok(())
+}
+
+/// Add or delete IPv4 addresses for eth0 (interface index 2)
+fn addr4_eth0(rtnl: &NlRouter, what: Rtm, addr: Ipv4Addr, prefix_len: u8) -> Result<()> {
+    let ifaddrmsg = IfaddrmsgBuilder::default()
+        .ifa_family(RtAddrFamily::Inet)
+        .ifa_prefixlen(prefix_len)
+        .ifa_scope(RtScope::Universe)
+        .ifa_index(2)
+        .rtattrs(RtBuffer::from_iter([
+            RtattrBuilder::default()
+                .rta_type(Ifa::Local)
+                .rta_payload(addr.octets().to_vec())
+                .build()?,
+            RtattrBuilder::default()
+                .rta_type(Ifa::Address)
+                .rta_payload(addr.octets().to_vec())
+                .build()?,
+        ]))
+        .build()?;
+
+    let _: NlRouterReceiverHandle<Rtm, Ifaddrmsg> = rtnl.send(
+        what,
+        NlmF::CREATE | NlmF::REQUEST,
+        NlPayload::Payload(ifaddrmsg),
+    )?;
+
+    Ok(())
+}
+
+/// Send DISCOVER with Rapid Commit, process ACK, configure address and route
+fn do_dhcp(rtnl: &NlRouter) -> Result<()> {
+    // Temporary link-local address and route avoid the need for raw sockets
+    route4_eth0(rtnl, Rtm::Newroute, Ipv4Addr::UNSPECIFIED)?;
+    addr4_eth0(rtnl, Rtm::Newaddr, Ipv4Addr::new(169, 254, 1, 1), 16)?;
+
+    // Send request (DHCPDISCOVER)
+    let socket = UdpSocket::bind("0.0.0.0:68").expect("Failed to bind");
+    let mut buf = [0; 576 /* RFC 2131, Section 2 */ ];
+
+    const REQUEST: &[u8; 300 /* From RFC 951: >= 60 B of options */ ] = const_str::concat_bytes!(
+        1, // BOOTREQUEST
+        0x01, // Hardware address type: Ethernet
+        6, // Hardware address length
+        0, // DHCP relay Hops
+        [1, 2, 3, 4], // Transaction ID: we're the only client, use a fixed one
+        [0, 0], // Seconds elapsed since beginning of acquisition or renewal
+        [0x80, 0x0], // DHCP message flags: Broadcast
+        [0; 16], // All-zero ciaddr, yiaddr, siaddr, giaddr
+        [0; 16], // Client hardware address, not set as we're the only client
+        [0; 64], // Server host name
+        [0; 128], // Boot file name
+        [0x63, 0x82, 0x53, 0x63], // Magic cookie prefix before options
+        // Options
+        [
+            53, 1, 1, // DHCPDISCOVER
+            80, 0, // Rapid Commit (RFC 4039)
+        ],
+        0xff, // End of options
+        [0; 54], // Pad (up to 300 bytes)
+    );
+
+    socket.set_broadcast(true)?;
+    socket.send_to(REQUEST, "255.255.255.255:67")?;
+
+    // Keep IPv6-only fast
+    let _ = socket.set_read_timeout(Some(Duration::from_millis(100)));
+
+    // Get and process response (DHCPACK) if any
+    if let Ok((len, _)) = socket.recv_from(&mut buf) {
+        let msg = &mut buf[..len];
+
+        let addr = Ipv4Addr::new(msg[16], msg[17], msg[18], msg[19]);
+        let mut netmask = Ipv4Addr::UNSPECIFIED;
+        let mut router = Ipv4Addr::UNSPECIFIED;
+        let mut mtu: u16 = 65520;
+        let mut p: usize = 240;
+        let mut resolv = fs::File::options()
+            .append(true)
+            .open("/etc/resolv.conf")
+            .context("Failed to open /etc/resolv.conf")?;
+
+        while p < len {
+            let o = msg[p];
+            let l: u8 = msg[p + 1];
+            p += 2; // Length doesn't include code and length field itself
+
+            if o == 1 {
+                // Option 1: Subnet Mask
+                netmask = Ipv4Addr::new(msg[p], msg[p + 1], msg[p + 2], msg[p + 3]);
+            } else if o == 3 {
+                // Option 3: Router
+                router = Ipv4Addr::new(msg[p], msg[p + 1], msg[p + 2], msg[p + 3]);
+            } else if o == 6 {
+                // Option 6: Domain Name Server
+                for dns_p in (p..p + l as usize).step_by(4) {
+                    let dns =
+                        Ipv4Addr::new(msg[dns_p], msg[dns_p + 1], msg[dns_p + 2], msg[dns_p + 3]);
+                    resolv
+                        .write_all(format!("nameserver {dns}\n").as_bytes())
+                        .context("Failed to write to resolv.conf")?;
+                }
+            } else if o == 26 {
+                // Option 26: Interface MTU
+                mtu = ((msg[p] as u16) << 8) | msg[p + 1] as u16;
+
+                // We don't know yet if IPv6 is available: don't go below 1280 B
+                mtu = mtu.clamp(1280, 65520);
+            } else if o == 0xff {
+                // Option 255: End (of options)
+                break;
+            }
+
+            p += l as usize;
+        }
+
+        let prefix_len: u8 = netmask.to_bits().leading_ones() as u8;
+
+        // Drop temporary address and route, configure what we got instead
+        route4_eth0(rtnl, Rtm::Delroute, Ipv4Addr::UNSPECIFIED)?;
+        addr4_eth0(rtnl, Rtm::Deladdr, Ipv4Addr::new(169, 254, 1, 1), 16)?;
+
+        addr4_eth0(rtnl, Rtm::Newaddr, addr, prefix_len)?;
+        route4_eth0(rtnl, Rtm::Newroute, router)?;
+
+        mtu_eth0(rtnl, mtu.into())?;
+    } else {
+        // Clean up: we're clearly too cool for IPv4
+        route4_eth0(rtnl, Rtm::Delroute, Ipv4Addr::UNSPECIFIED)?;
+        addr4_eth0(rtnl, Rtm::Deladdr, Ipv4Addr::new(169, 254, 1, 1), 16)?;
+    }
+
+    Ok(())
+}
+
+/// Wait for SLAAC to complete or fail
+fn wait_for_slaac(rtnl: &NlRouter) -> Result<()> {
+    let mut global_seen = false;
+    let mut global_wait = true;
+    let mut ll_seen = false;
+
+    // Busy-netlink-loop until we see a link-local address, and a global unicast
+    // address as long as we might expect one (see below)
+    while !ll_seen || (global_wait && !global_seen) {
+        let ifaddrmsg = IfaddrmsgBuilder::default()
+            .ifa_family(RtAddrFamily::Inet6)
+            .ifa_prefixlen(0)
+            .ifa_scope(RtScope::Universe)
+            .ifa_index(2)
+            .build()?;
+
+        let recv = rtnl.send(Rtm::Getaddr, NlmF::ROOT, NlPayload::Payload(ifaddrmsg))?;
+
+        for response in recv {
+            let header: Nlmsghdr<Rtm, Ifaddrmsg> = response?;
+            if let NlPayload::Payload(p) = header.nl_payload() {
+                if p.ifa_scope() == &RtScope::Link {
+                    // A non-tentative link-local address implies we sent a
+                    // router solicitation that didn't get any response
+                    // (IPv4-only)? Stop waiting for the router in that case
+                    if *p.ifa_flags() & IfaF::TENTATIVE != IfaF::TENTATIVE {
+                        global_wait = false;
+                    }
+
+                    ll_seen = true;
+                } else if p.ifa_scope() == &RtScope::Universe {
+                    global_seen = true;
+                }
+            }
+        }
+    }
+
+    Ok(())
+}
 
 pub fn configure_network() -> Result<()> {
     // Allow unprivileged users to use ping, as most distros do by default.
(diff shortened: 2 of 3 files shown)


sbrivio-rh:
Supersedes https://github.com/AsahiLinux/muvm/pull/64/. Test passt static builds, along with RPMs and Debian packages (x86 only, sorry) at https://passt.top/builds/latest/x86_64/

crates/muvm/Cargo.toml
```diff
@@ -15,6 +15,7 @@ byteorder = { version = "1.5.0", default-features = false, features = ["std"] }
 env_logger = { version = "0.11.3", default-features = false, features = ["auto-color", "humantime", "unstable-kv"] }
 krun-sys = { path = "../krun-sys", version = "1.9.1", default-features = false, features = [] }
 log = { version = "0.4.21", default-features = false, features = ["kv"] }
+neli = { version = "0.7.0-rc2" }
```
teohhanhui, line 18:
```suggestion
neli = { version = "0.7.0-rc2", default-features = false, features = [] }
```

sbrivio-rh: Ah, thanks, I had no idea about the whole "default features" mechanism.

crates/muvm/src/guest/net.rs
```diff
@@ -33,63 +116,135 @@ pub fn configure_network() -> Result<()> {
...
-                    "`dhcpcd` process terminated by signal: {}",
-                    output
-                        .status
-                        .signal()
-                        .expect("either one of status code or signal should be set")
-                )
-            };
-            Err(err)?;
-        }
+    let (rtnl, _) = NlRouter::connect(NlFamily::Route, None, Groups::empty())?;
```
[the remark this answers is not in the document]
sbrivio-rh: Should I create some intermediate functions? I thought I already abstracted enough by separating those netlink operations. I could probably move the DHCP stuff to its own function and the post-SLAAC stuff as well, but it's not really practical to have a separate "SLAAC" function because some steps need to be done before sending DHCP requests.

To me it's actually easier to follow if it's sequential/flattened, because adding and removing routes and addresses is intertwined with some other operations, but I guess I see the case for intermediate functions (even though I don't find anything wrong with a configure_network() function setting interface flags).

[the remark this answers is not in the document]
teohhanhui: All fair points. I also favour not making unnecessary abstractions.

sbrivio-rh:
Hmm, I can't test this anymore after rebasing to latest upstream. I'm getting one of these two errors ("Failed to create the microVM" about 30% of the times, Failed to execute `muvm-server` as child process about 70%). These are attempts in a relatively tight loop:

```
$ ./target/debug/muvm -- /bin/false
Error: Failed to execute `muvm-server` as child process

Caused by:
    No such file or directory (os error 2)
$ ./target/debug/muvm -- /bin/false
Error: Failed to create the microVM

Caused by:
    Invalid argument (os error 22)
$ ./target/debug/muvm -- /bin/false
Error: Failed to create the microVM

Caused by:
    Invalid argument (os error 22)
$ ./target/debug/muvm -- /bin/false
Error: Failed to execute `muvm-server` as child process

Caused by:
    No such file or directory (os error 2)
$ ./target/debug/muvm -- /bin/false
Error: Failed to execute `muvm-server` as child process

Caused by:
    No such file or directory (os error 2)
```

With `RUST_LOG=debug`:

```
[2024-11-25T19:05:58Z ERROR krun] Building the microVM failed: GuestMemoryMmap(UnsortedMemoryRegions)
```

or:

```
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::linux::passthrough] do_lookup: "muvm-server"
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::server] opcode: 3
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::worker] Fs: queue event: 1
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::server] opcode: 1
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::linux::passthrough] do_lookup: "muvm-server"
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::worker] Fs: queue event: 1
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::server] opcode: 1
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::linux::passthrough] do_lookup: "muvm-server"
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::worker] Fs: queue event: 1
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::server] opcode: 15
[2024-11-25T19:08:04Z DEBUG devices::virtio::fs::linux::passthrough] read: 56
Error: [2024-11-25T19:08:04Z DEBUG devices::virtio::fs::server] opcode: 25
Failed to execute `muvm-server` as child process
```

I reverted a few commits but I can't seem to get this to work anymore.

crates/muvm/src/guest/net.rs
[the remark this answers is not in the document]
sbrivio-rh: I actually tried splitting stuff in two further functions, it looks okay I guess... I'll post that too once I figure out why muvm just stopped working entirely for me (even without my changes).

sbrivio-rh:
Oh, okay, I didn't pull for a while. It works if I do:
```
284b520 (HEAD) Revert "Share /dev/shm as a separate mount with DAX"
44ef0b6 Revert "Protect muvm-server with a cookie"
cdf7027 Revert "Start a privileged muvm-server"
df6c549 Revert "Implement a host memory monitor"
```

...posting another version of that commit with two functions taken out of configure_network() now that I can test things. Those other issues I'm facing, I have no idea where to start debugging them...

sbrivio-rh:
> ```
> $ ./target/debug/muvm -- /bin/false
> Error: Failed to execute `muvm-server` as child process
> 
> Caused by:
>     No such file or directory (os error 2)
> ```

Fixed by https://github.com/AsahiLinux/muvm/pull/112

> ```
> $ ./target/debug/muvm -- /bin/false
> Error: Failed to create the microVM
> 
> Caused by:
>     Invalid argument (os error 22)
> ```

Fixed by updating libkrun.

sbrivio-rh:
This is now supported by passt [2024_11_27.c0fbc7e](https://archives.passt.top/passt-user/20241127142126.3c53066e@elisabeth/), matching Fedora updates [passt-0^20241127.gc0fbc7e-1.fc40](https://bodhi.fedoraproject.org/updates/FEDORA-2024-c6a6e8bce8),  [passt-0^20241127.gc0fbc7e-1.fc41](https://bodhi.fedoraproject.org/updates/FEDORA-2024-ae2aa0162c), [passt-0^20241127.gc0fbc7e-1.fc42](https://bodhi.fedoraproject.org/updates/FEDORA-2024-c0a13afb0e), as well as Debian's [passt-0.0~git20241127.c0fbc7e-1](https://tracker.debian.org/news/1588650/accepted-passt-00git20241127c0fbc7e-1-source-into-unstable/).

slp:
Thanks a lot @sbrivio-rh , I really like this approach. A couple questions:

- What should we do with `resolv.conf`? Perhaps I'm missing it, but doesn't seem like it's doing nameserver resolution.
- This introduces a dependency on neli, but it's not yet packaged into Fedora. Do you want to package it yourself?

sbrivio-rh:
> * What should we do with `resolv.conf`? Perhaps I'm missing it, but doesn't seem like it's doing nameserver resolution.

Ah, right, I thought `resolv.conf` could simply be the one from the host, but with systemd-resolved it's common to have a loopback address as resolver, and that wouldn't work.

So, yes, I should also configure `resolv.conf` here from DHCP options (6 and 119) and NDP (RDNSS, option 25), passt already sends the right ones. Let me fix this.

> * This introduces a dependency on neli, but it's not yet packaged into Fedora. Do you want to package it yourself?

Oops, I didn't check. I'm not exactly the right person as I barely understand what a crate is (do I?) and I don't even use Fedora regularly, but yes, I can probably do that, it should be low effort.

I found this abandoned Copr by the way, https://copr.fedorainfracloud.org/coprs/zurdo/i3status-rs-update/package/rust-neli/, it looks pretty easy. Let me give it a try, but if you know somebody else who could be interested...

sbrivio-rh:
> This introduces a dependency on neli, but it's not yet packaged into Fedora. Do you want to package it yourself?

Naive question: if it's statically linked, does it really become a dependency? Or is it just a build dependency? Does that matter also if the crate is downloaded as needed...?

sbrivio-rh:
> It's just a build dependency. In Fedora, every crate you depend on must be independently packaged, and builds are done offline. Luckily, rust2rpm helps a lot with this.

Oh, oops, I just had a look at https://src.fedoraproject.org/user/slp/projects... let me package that. :)

crates/muvm/src/guest/net.rs
```diff
@@ -1,14 +1,215 @@
...
+                                        msg[p + 4], msg[p + 5]);
+            } else if o == 0xff { // Option 255: End (of options)
+                break;
+            }
+
+            l += 2; // Length doesn't include code and length field itself
+            p += l as usize;
+        }
+
+        let prefix_len : u8 = netmask.to_bits().leading_ones() as u8;
```
teohhanhui, line 158: I was wondering why the bump to Rust 1.80, and this is the reason...

https://doc.rust-lang.org/stable/core/net/struct.Ipv4Addr.html#method.to_bits

(No action necessary here.)

sbrivio-rh: Correct :)

sbrivio-rh:
> So, yes, I should also configure `resolv.conf` here from DHCP options (6 and 119) and NDP (RDNSS, option 25), passt already sends the right ones. Let me fix this.

I just added support for nameservers over DHCP (option 6), omitting for the moment:
- handling of the domain search list (option 119) also implemented by passt. It's at least 10-20 LoCs due to domain compression encoding from RFC 1035, and I doubt it's of any use for muvm... we can add it later for completeness anyway but I'd rather fix the current situation first
- handling of NDP option 25 (RDNSS), implemented by passt as well, for DNS resolvers over SLAAC. That's also some extra bit of complexity. I think this one makes sense no matter what, for IPv6-only setups, but I'd rather avoid to make this patch explode at the moment

sbrivio-rh:
Oops, I just noticed the `cargo fmt -- --check` warnings, I thought `cargo clippy` would be enough. Fixing those as well...

sbrivio-rh:
> Oops, I just noticed the `cargo fmt -- --check` warnings, I thought `cargo clippy` would be enough. Fixing those as well...

Gosh, the reformatted version looks horrible, with 100 columns that don't fit pretty much anywhere and things wildly misaligned. Is `cargo fmt` enforced?

crates/muvm/src/guest/net.rs
```diff
@@ -1,14 +1,224 @@
...
+fn do_dhcp(rtnl: &NlRouter) -> Result<()> {
+    // Temporary link-local address and route avoid the need for raw sockets
+    route4_eth0(rtnl, Rtm::Newroute, Ipv4Addr::UNSPECIFIED)?;
+    addr4_eth0(rtnl, Rtm::Newaddr, Ipv4Addr::new(169, 254, 1, 1), 16)?;
+
+    // Send request (DHCPDISCOVER)
+    let socket = UdpSocket::bind("0.0.0.0:68").expect("Failed to bind");
+    let mut buf = [0; 576 /* RFC 2131, Section 2 */ ];
+
+    const REQUEST: [u8; 300 /* From RFC 951: >= 60 B of options */ ] = [
```
teohhanhui, line 106: I have a suggestion: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b68067b21b871a7d80e6d9a81d564e92

sbrivio-rh: How would you initialise bytes after the `sname` and `file` fields, though? That is, the DHCP cookie and the options.

Besides, if you prefer a version with everything on its own line, a struct with `#[repr(packed)]` could be used instead (at least up to the options... those would need a different description). [This](https://passt.top/passt/tree/dhcp.c?id=020c8b7127e38872e68bffb30ad388001e088552#n86), for example, is how passt does it in C.

But my idea here was, given that we we don't have to be flexible at all, to keep it in sixteen lines (because it's possible), so that you have an immediate overview of the whole thing without scrolling back and forth.

sbrivio-rh:
> > It's just a build dependency. In Fedora, every crate you depend on must be independently packaged, and builds are done offline. Luckily, rust2rpm helps a lot with this.
> 
> Oh, oops, I just had a look at https://src.fedoraproject.org/user/slp/projects... let me package that. :)

Package reviews:
https://bugzilla.redhat.com/show_bug.cgi?id=2329411
https://bugzilla.redhat.com/show_bug.cgi?id=2329412

crates/muvm/src/guest/net.rs
teohhanhui: I consider the struct to be an overkill like you said, but I also think the original formatting you had was unreadable because they're all packed together. I think my suggestion allows easily seeing what's at byte N.

sbrivio-rh: But still, I don't understand: how would you initialise the cookie and option bytes? That is, these ones:

```
        0x63, 0x82, 0x53, 0x63, /* DHCP (magic) cookie, then options: */
        53, 1, 1 /* DISCOVER */, 80, 0 /* Rapid commit */, 0xff, // Done
```

...they come after a lot of zeroes. At the beginning, I tried something similar to your suggestion, but instead of ending it with `r`, I tried to include a given number of zeroes with [0u8; n] in the middle of the initialiser. But it's not supported, and nothing like that seems to be supported from what I read.

> I think my suggestion allows easily seeing what's at byte N.

It's actually not so important for DHCP, because fields are not aligned by design. Making it a bit artistic and all squared looked easier on my eyes to me, which makes me wonder: can it be that our editors have different setting so what's readable for me looks messy to you? This is how it looks like for me, in the editors I use:

geany

joe

teohhanhui: Readability is subjective, of course. A sea of zeroes interspersed with some values and comments is hard to follow IMO.

I have some ideas on what could work. Will give it a try later.

sbrivio-rh: > A sea of zeroes interspersed with some values and comments is hard to follow IMO.

but... but... art :cry:

teohhanhui: There's [`concat_bytes!`](https://doc.rust-lang.org/std/macro.concat_bytes.html) but it's still unstable: https://github.com/rust-lang/rust/issues/87555

I think we can use [`const_str::concat_bytes!`](https://docs.rs/const-str/latest/const_str/macro.concat_bytes.html) in the meantime.

teohhanhui: https://github.com/sbrivio-rh/muvm/pull/1

sbrivio-rh: I really appreciate the effort, and the creative usage of `concat_bytes!`, I just can't refrain from noticing that a compact, four-byte aligned, sixteen-line representation of a simple UDP datagram is now thirty-six lines, more than double, with a lot of RFC-951-derived cruft (which nobody cares about) spelled out prominently.

It's harder to get a quick overview of the options (the only thing that really matters here).

> Readability is subjective, of course

...except that your way is better? I don't get it.

Well, anyway, if `cargo fmt` is enforced, or in any case if this is the only alternative to avoid screen-scraping from passt (and getting the thing to start fast), so be it.

Where do I click? Should I accept your pull request on my tree and then force-push here? I have to admit that while I've been using git for 19 years, I'm kind of lost with GitHub. I would need to update the comments, too.

By the way, is there a way to calculate the 54 at the end using something equivalent to C's `300 - offsetof(struct msg, end)`?

teohhanhui: I'll try to use line-end comments instead if that's preferable.

> By the way, is there a way to calculate the 54 at the end using something equivalent to C's 300 - offsetof(struct msg, end)?

I'll define `REQUEST_LEN` :)

teohhanhui: > ...except that your way is better? I don't get it.

I'm sorry that it came off that way. What I meant is that there's no way for us to judge what's the right way here, except that yes, the formatting you had definitely trips up `rustfmt`, and even if we were to add an ignore here, it'd still look out of place with the rest of the codebase...

sbrivio-rh: > I'll try to use line-end comments instead if that's preferable.

...I don't think it would make a big difference, I still see it as an explosion of verbosity for the sake of pretending to document details which happen to come from the mid 80s and which are not relevant anymore.

> I'll define `REQUEST_LEN` :)

It's not so much about the "300" itself, it's about the`offsetof` equivalent to calculate the padding. I couldn't find any.

teohhanhui: I found https://doc.rust-lang.org/core/mem/macro.offset_of.html but would it help in our case, considering we only have an array?

teohhanhui: Updated. Now 21 lines vs. your original 17 lines. I hope it's not too inelegant? :bow:

sbrivio-rh: > I found https://doc.rust-lang.org/core/mem/macro.offset_of.html but would it help in our case, considering we only have an array?

Aaah, right, and turning that into a struct makes it quite verbose again. Never mind, I don't think we'll ever need to change those.

> Updated. Now 21 lines vs. your original 17 lines. I hope it's not too inelegant? 🙇

Thanks, much much better. I still cry a bit inside because of what happened to my square packet, but okay, if `rustfmt` says so...

I'll now carefully click and try to merge things back to here...

sbrivio-rh: Not carefully enough, now I have that dangling Cargo.toml update from you... grr. I'm trying to rebuild a sane situation but it's taking a bit.

sbrivio-rh: Pushed, I hope it's fine now...

But really, what do you all do with Cargo.toml if you can't `gitignore` it?

teohhanhui: Do you mean `Cargo.lock`? Just `git add` it after you run `cargo build` and it should be correct.

sbrivio-rh: Right, sorry, `Cargo.lock`. So I'm actually supposed to commit it, even it's automatically generated...?

sbrivio-rh: Yes, I read that proposal (or is it a recommendation? It's not stated...) (_after_ proposing the `.gitignore` change, oops). I'm still not convinced it's worth the madness as I never had issues of that sort (well, with C, at least). Anyway, pushing those changes as well...

teohhanhui (approved):

crates/muvm/src/guest/net.rs
teohhanhui: RE: art

https://mastodon.social/@richlv/113567941242044603 :crying_cat_face:

slp:
LGTM, but please let's wait to merge this one until the packaging stuff is solved, so we aren't blocked on making releases.

teohhanhui:
> packaging stuff

Now we have one more crate to package: `const_str` :laughing:

sbrivio-rh:
> > packaging stuff
> 
> Now we have one more crate to package: `const_str` 😆

Gosh. I can try to do it as well, but that will take even longer... is it really worth it? Should we consider **temporarily** going back to my original version that didn't use it?

teohhanhui:
> Gosh. I can try to do it as well, but that will take even longer... is it really worth it? Should we consider temporarily going back to my original version that didn't use it?

There's another way. We could create a `muvm_macros` crate and put `concat_bytes` in there?

sbrivio-rh:
> There's another way. We could create a `muvm_macros` crate and put `concat_bytes` in there?

I don't know what the difference would be. I'm trying to build a spec file for const_str meanwhile, currently struggling with the license file.

teohhanhui:
> I don't know what the difference would be.

I mean copying the code into our own crate, if the license permits... I think that means we don't have to go through package review because it won't be a separate package at all?

sbrivio-rh:
> I think that means we don't have to go through package review because it won't be a separate package at all?

True but...

> I mean copying the code into our own crate

from a maintainability perspective, I'd say that screen-scraping the output of passt (https://github.com/AsahiLinux/muvm/pull/64/) is actually preferable to that.

slp:
> Well, current spec file and corresponding error:
> 
> https://hastebin.skyra.pw/hudaqicari.prolog

We had this problem with some of the rust-vmm packages. If the LICENSE file is in the root of the workspace, it doesn't get included in the crate. The trick here is adding the file as a source as done here:

https://src.fedoraproject.org/rpms/rust-virtio-queue/blob/rawhide/f/rust-virtio-queue.spec

sbrivio-rh:
> > Well, current spec file and corresponding error:
> > https://hastebin.skyra.pw/hudaqicari.prolog
> 
> We had this problem with some of the rust-vmm packages. If the LICENSE file is in the root of the workspace, it doesn't get included in the crate. The trick here is adding the file as a source as done here:
> 
> https://src.fedoraproject.org/rpms/rust-virtio-queue/blob/rawhide/f/rust-virtio-queue.spec

That's what I did there (I think?):

```
Source1:        https://raw.githubusercontent.com/Nugine/const-str/00ce0d143164e1e393a5dcdda6ab2f0dbae9a243/LICENSE

[...]

%license %{crate_instdir}/LICENSE
```

...there must be some subtle detail I'm missing.

sbrivio-rh:
> That's what I did there (I think?):
> 
> ```
> Source1:        https://raw.githubusercontent.com/Nugine/const-str/00ce0d143164e1e393a5dcdda6ab2f0dbae9a243/LICENSE
> 
> [...]
> 
> %license %{crate_instdir}/LICENSE
> ```
> 
> ...there must be some subtle detail I'm missing.

Grr, of course. From the rust-virtio-queue spec file:

```
%prep
%autosetup -n %{crate}-%{version} -p1
%cargo_prep
cp %{SOURCE1} .
cp %{SOURCE2} .
```

Thanks @slp for the example.

sbrivio-rh:
Fedora package review request for `rust-const-str` at: https://bugzilla.redhat.com/show_bug.cgi?id=2330150

teohhanhui:
Looks like the merge conflicts have not been resolved... :see_no_evil:

sbrivio-rh:
> Looks like the merge conflicts have not been resolved... 🙈

I just clicked around randomly, I'm anyway not able to test it anymore because of https://github.com/AsahiLinux/muvm/pull/117#issuecomment-2567922524 and a new `Error: Address already in use (os error 98)` coming out of nowhere (no `EADDRINUSE` from strace, and extra difficult to bisect because of that commit I need to keep reverted plus Cargo.toml committed in the repository... now writing some scripts to handle that at least for my git sanity).

I'll push fixed conflicts once I can test this again.

sbrivio-rh:
> and a new `Error: Address already in use (os error 98)` coming out of nowhere

Oh, that's because my usual trick of commenting out `setgid(gid).context("Failed to setgid")?;` and `setuid(uid).context("Failed to setuid")?;` in `setup_user()` to get a root shell conveniently (I need to do stuff with iproute2 in the guest) doesn't work anymore. The second instance of `muvm-server` fails to start.

On the other hand, if I skip starting it, then everything hangs...

WhatAmISupposedToPutHere:
> > Looks like the merge conflicts have not been resolved... 🙈
> 
> I just clicked around randomly, I'm anyway not able to test it anymore because of [#117 (comment)](https://github.com/AsahiLinux/muvm/pull/117#issuecomment-2567922524) and a new `Error: Address already in use (os error 98)` coming out of nowhere (no `EADDRINUSE` from strace, and extra difficult to bisect because of that commit I need to keep reverted plus Cargo.lock committed in the repository... now writing some scripts to handle that at least for my git sanity).
> 
> I'll push fixed conflicts once I can test this again.

Try using --sommelier, it might on a headle

sbrivio-rh:
> Try using --sommelier, it might on a headless machine that way

...thanks, yes, that also works because it reverts the effect of commit `b40bf03d2ce4 ("Default to using x11bridge instead of sommelier"), and I guess that the sommelier path has a proper fallback path (I didn't really check): I don't have sommelier installed.

On the other hand I guess we should have the same fallback with no options at all (at least as long as muvm is supposed to be a generic tool).

slp:
> Fedora package review request for `rust-const-str` at: https://bugzilla.redhat.com/show_bug.cgi?id=2330150

Looks like the package review got stuck because it depends on packaging `rust-const-str-proc-macro` first (same repo, different crate). Do you need a hand with that?

sbrivio-rh:
> > Fedora package review request for `rust-const-str` at: https://bugzilla.redhat.com/show_bug.cgi?id=2330150
> 
> Looks like the package review got stuck because it depends on packaging `rust-const-str-proc-macro` first (same repo, different crate). Do you need a hand with that?

Thanks for the offer! But actually, this is stuck because I'm waiting for https://github.com/jbaublitz/neli/pull/241. It's [not blocking](https://bugzilla.redhat.com/show_bug.cgi?id=2329411#c9), but I thought it makes sense to wait a bit anyway as the author is quite close to merge it. Once it's merged, I would then spend some "hours of fun" on packaging matters (not just that package) and try to sort all those.

List of things (kind of) blocking this at the moment:
- I can't just run `muvm` anymore on my system (workaround: `--sommelier`... but it's not great), see https://github.com/AsahiLinux/muvm/pull/111#issuecomment-2568116194
- I can't simply comment out the setuid()/setgid() calls anymore to get a guest root shell. That would be pretty convenient to test the whole thing, see https://github.com/AsahiLinux/muvm/pull/111#issuecomment-2568041413. Not really blocking either
- `rust-const-str`: trivial issues reported in review (so far), https://bugzilla.redhat.com/show_bug.cgi?id=2330150. I plan to "respin" once I get back to packaging stuff
- `rust-neli-proc-macros`: package review at https://bugzilla.redhat.com/show_bug.cgi?id=2329411, waiting for https://github.com/jbaublitz/neli/pull/241
- `rust-neli`: package review at https://bugzilla.redhat.com/show_bug.cgi?id=2329412, waiting for `rust-neli-proc-macros`

WhatAmISupposedToPutHere:
> * I can't simply comment out the setuid()/setgid() calls anymore to get a guest root shell. That would be pretty convenient to test the whole thing, see

There is now an "official" way to getting the root shell, by running sth like `muvm -p 3335 -it bash` from a separate terminal tab.

sbrivio-rh:
> There is now an "official" way to getting the root shell, by running sth like `muvm -p 3335 -it bash` from a separate terminal tab.

Ah, thanks, I didn't know that. For some reason it hangs for a while, then aborts for me:
```
$ target/debug/muvm -p 3335 -it bash
Error: could not request launch to server: could not connect to muvm server: Connection reset by peer (os error 104)
```
with the guest started as `target/debug/muvm --sommelier bash`.

On the other hand, it's much less usable for me than something like `muvm -r / --root` (you know, I'm using a terminal for it already...), so I'd rather invest a moment into implementing that if I ever find the time, rather than debugging this.

sbrivio-rh:
> > > Fedora package review request for `rust-const-str` at: https://bugzilla.redhat.com/show_bug.cgi?id=2330150
> > 
> > 
> > Looks like the package review got stuck because it depends on packaging `rust-const-str-proc-macro` first (same repo, different crate). Do you need a hand with that?
> 
> Thanks for the offer! But actually, this is stuck because I'm waiting for [jbaublitz/neli#241](https://github.com/jbaublitz/neli/pull/241). It's [not blocking](https://bugzilla.redhat.com/show_bug.cgi?id=2329411#c9), but I thought it makes sense to wait a bit anyway as the author is quite close to merge it. Once it's merged, I would then spend some "hours of fun" on packaging matters (not just that package) and try to sort all those.
> 
> List of things (kind of) blocking this at the moment:
> 
>     * I can't just run `muvm` anymore on my system (workaround: `--sommelier`... but it's not great), see [guest/net: New implementation of network setup with SLAAC and own DHC… #111 (comment)](https://github.com/AsahiLinux/muvm/pull/111#issuecomment-2568116194)
> 
>     * I can't simply comment out the setuid()/setgid() calls anymore to get a guest root shell. That would be pretty convenient to test the whole thing, see [guest/net: New implementation of network setup with SLAAC and own DHC… #111 (comment)](https://github.com/AsahiLinux/muvm/pull/111#issuecomment-2568041413). Not really blocking either
> 
>     * `rust-const-str`: trivial issues reported in review (so far), https://bugzilla.redhat.com/show_bug.cgi?id=2330150. I plan to "respin" once I get back to packaging stuff
> 
>     * `rust-neli-proc-macros`: package review at https://bugzilla.redhat.com/show_bug.cgi?id=2329411, waiting for [Migrate neli-proc-macros to syn 2.0.x jbaublitz/neli#241](https://github.com/jbaublitz/neli/pull/241)

Now superseded by https://github.com/jbaublitz/neli/pull/256 (merged), so I can proceed with everything. I'll continue as soon as I find a couple of hours in a row.

sbrivio-rh:
Now it should be all clean again. I also added handling of option 26 (Interface MTU), as passt can conveniently use 65520 bytes (the host kernel does segmentation for TCP anyway). That's one small step for a DHCP client, one giant leap for throughput.

I can't fathom for the life of me how you all can live without version logs for patches.

sbrivio-rh:
> * `rust-neli-proc-macros`: package review at https://bugzilla.redhat.com/show_bug.cgi?id=2329411, waiting for [Migrate neli-proc-macros to syn 2.0.x jbaublitz/neli#241](https://github.com/jbaublitz/neli/pull/241)

I sent a new version for (package) review. Now we have an additional problem though:

> * `rust-neli`: package review at https://bugzilla.redhat.com/show_bug.cgi?id=2329412, waiting for `rust-neli-proc-macros`

The updated version, with https://github.com/jbaublitz/neli/pull/182, needs `derive-builder` 0.11. Fedora packages 0.20.2. I'll pretend I know what I'm doing and send a patch for https://github.com/jbaublitz/neli/, but generally speaking this doesn't seem like a viable approach to development.

sbrivio-rh:
> * `rust-const-str`: trivial issues reported in review (so far), https://bugzilla.redhat.com/show_bug.cgi?id=2330150. I plan to "respin" once I get back to packaging stuff

Additional review request: `rust-const-str-proc-macro`, https://bugzilla.redhat.com/show_bug.cgi?id=2338679.

teohhanhui:
@sbrivio-rh It's specified by Fedora's Rust packaging guidelines: https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_packaging_multiple_versions

sbrivio-rh:
> @sbrivio-rh It's specified by Fedora's Rust packaging guidelines: https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_packaging_multiple_versions

Sure, that makes sense, I just think that having a dependency manager and in general an ecosystem happily pulling in whatever new crate du jour (that's just Rust and it's all fine per se) along with the requirement that every and each dependency crate is packaged separately (and that's Fedora's policy) poses a significant barrier for contributors (especially occasional ones), and not in a good way, without technical merit or quality benefit whatsoever. Cargo already enforces what's needed.

This thing was pretty much ready three months ago. But okay, sure, also `libnl` needed to be packaged for all distributions eventually. I hope we're almost done adding packages (for this).

sbrivio-rh:
Current status of package dependencies in Fedora:
- I'm now the proud maintainer of the [rust-neli-proc-macros](https://src.fedoraproject.org/rpms/rust-neli-proc-macros) package
- rust-neli: package ready for review at https://bugzilla.redhat.com/show_bug.cgi?id=2329412, no reviewers for the moment
- rust-const-str-proc-macro: ready for re-re-review at https://bugzilla.redhat.com/show_bug.cgi?id=2338679, @decathorpe currently assigned
- rust-const-str: review at https://bugzilla.redhat.com/show_bug.cgi?id=2330150, but it needs to wait for rust-const-str-proc-macros

decathorpe:
I'll be taking another look, as soon as I'm home from FOSDEM.

sbrivio-rh:
rust-const-str-proc-macro packaged, current status of dependencies in Fedora:

* I'm now the proud maintainer of the [rust-neli-proc-macros](https://src.fedoraproject.org/rpms/rust-neli-proc-macros) and of the [rust-const-str-proc-macro](https://src.fedoraproject.org/rpms/rust-const-str-proc-macro) packages

* rust-neli: package ready for review at https://bugzilla.redhat.com/show_bug.cgi?id=2329412, no reviewers for the moment

* rust-const-str: now that rust-const-str-proc-macro is packaged, I can proceed with this as soon as the latter reaches mirrors, then I'll get back to the review request at https://bugzilla.redhat.com/show_bug.cgi?id=2330150

sbrivio-rh:
> Looks like both packages have finally been approved. 🥳

Right! Update:

- I'm now the proud maintainer of the [rust-const-str](https://src.fedoraproject.org/rpms/rust-const-str) package as well, builds are available
- I'm also the proud maintainer of the [rust-neli](https://src.fedoraproject.org/rpms/rust-neli) package, but unfortunately tests fail on s390x, so builds are not available yet. From: https://kojipkgs.fedoraproject.org//work/tasks/7795/130447795/build.log
```
test rtnl::test::test_rta_deserialize ... FAILED
```
I'm currently looking into endianness issues.

sbrivio-rh:
https://github.com/jbaublitz/neli/issues/273 -- feel free to contribute if you have some ideas...

decathorpe:
Sorry about that, usually build failures like that should be caught during package review. My bad :(

sbrivio-rh:
> Sorry about that, usually build failures like that should be caught during package review. My bad :(

It's not really a build failure, it's a test failure on big-endian only. How would you catch that during review?

decathorpe:
> > Sorry about that, usually build failures like that should be caught during package review. My bad :(
> 
> It's not really a build failure, it's a test failure on big-endian only. How would you catch that during review?

Well, a test failure *is* a build, failure, as you've found out. :) Many people (and the fedora-create-review tool, by default) run a scratch build for all architectures when submitting a package for review, which would have caught this.

sbrivio-rh:
So, after further baby-sitting from @decathorpe, I can now finally announce that we have builds (at least for Fedora Rawhide) of all the dependencies added here:
- [rust-const-str-proc-macro-0.6.2-1.fc43](https://bodhi.fedoraproject.org/updates/FEDORA-2025-494cad7668)
- [rust-const-str-0.6.2-1.fc43](https://bodhi.fedoraproject.org/updates/FEDORA-2025-8e5455971c)
- [rust-neli-proc-macros-0.2.0~rc3-1.fc43](https://bodhi.fedoraproject.org/updates/FEDORA-2025-e65d41a361)
- [rust-neli-0.7.0~rc3-2.fc43](https://bodhi.fedoraproject.org/updates/FEDORA-2025-716ae39cd4)

@teohhanhui @slp is that enough to merge this, or do we have to wait for packages for a specific Fedora version...?

teohhanhui:
@sbrivio-rh I think this needs a rebase, so that the CI job can have a fresh run.

sbrivio-rh:
Oops, I saw "No conflicts with base branch" and I thought everything was fine, but now that I'm rebasing I actually see conflicts...

sbrivio-rh:
New pull, new warnings:
```
$ ./target/debug/muvm bash
Error setting up FEX in binfmt_misc: Failed to find `FEXInterpreter` in PATH
Failed to find or configure FEX, falling back to Box
Error setting up Box in binfmt_misc: Failed to find `box64` in PATH
No emulators were configured, x86 emulation may not work
Using default interface naming scheme 'v255'.
```

anyway, those are just warnings I'll ignore for now. Things seem to work.

sbrivio-rh:
Reverting 5f9f9d9ecf7e62769efd9078a9f355ba294f150b fixes those, by the way.

nrabulinski:
From my understanding this is the expected behavior, those warnings will go away once I get around implementing #159. Trying to configure a binfmt_misc handler is a good default for now since the bulk of muvm usage comes from folks running Asahi kernel wanting to run 4K page apps.

sbrivio-rh:
> those warnings will go away once I get around implementing https://github.com/AsahiLinux/muvm/issues/159

It looks like that, yes, I wasn't aware of it, thanks for the pointer and thanks in advance for the implementation.

sbrivio-rh:
All the dependencies are now available in "stable" repositories for Fedora 40, 41, 42 and Rawhide. Is there anything else blocking this?

PR #114, opened 2024-11-28
muvm-guest: For debug builds, source muvm-hidpipe from current path a…

sbrivio-rh:
…s well

Same as commit d89ddbe7d022 ("env: For debug builds, source muvm binaries from current executable path"), but for muvm-hidpipe, which was imported into muvm by commit 44d33a39eb20 ("hidpipe: merge hidpipe project into muvm").

Otherwise I get:

  Error: Failed to execute `muvm-hidpipe` as child process

  Caused by:
      No such file or directory (os error 2)

### DIFF
--- a/crates/muvm/src/guest/bin/muvm-guest.rs
+++ b/crates/muvm/src/guest/bin/muvm-guest.rs
@@ -57,7 +57,8 @@ fn main() -> Result<()> {
 
     configure_network()?;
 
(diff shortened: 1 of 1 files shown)


sbrivio-rh:
Cc: @teohhanhui @slp

teohhanhui (approved):

PR #117, opened 2024-12-04
Make x11bridge the default mechanism for exposing graphics to the guest

slp:


asahilina (approved):

WhatAmISupposedToPutHere (approved):

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -81,12 +80,7 @@ fn main() -> Result<()> {
     let pulse_path = pulse_path.join("native");
     setup_socket_proxy(pulse_path, 3333)?;
 
-    #[cfg(feature = "x11bridge")]
-    let direct_x11 = setup_x11_forwarding(run_path)?;
-    #[cfg(not(feature = "x11bridge"))]
-    let direct_x11 = false;
-
-    if !direct_x11 {
+    if !setup_x11_forwarding(run_path)? {
```
teohhanhui, line 83: I think this function shouldn't return bool, but just propagate the error if something is wrong?

asahilina: It returns `Result<bool>`. `Err(x)` means something is wrong, `Ok(true)` means X11 forwarding is enabled, `Ok(false)` means X11 forwarding is disabled (and sommelier should be used instead).

teohhanhui: Yes, I understand. But I think it should rather just error instead of falling back to sommelier automatically? The user can be expected to try with `--sommelier` flag if the default doesn't work...

teohhanhui: Hmm... I need to look at the code again. Do we have an actual way to tell whether x11bridge is explicitly disabled (i.e. passing the `--sommelier` flag), vs. the case of it just failing up be set up? I think the latter should result in an error instead.

teohhanhui: Uhh... Never mind. `HOST_DISPLAY` is only set when `--sommelier` flag is not being passed.

sbrivio-rh:
This breaks my typical usage (headless, on a server, no X11/Wayland, no DRI):

```
$ ./target/debug/muvm sh
Error: X11 forwarding requested but DISPLAY is unset

Caused by:
    environment variable not found
```

PR #120, opened 2024-12-08
env: Force a separate Firefox profile

hoshinolina:
Otherwise Firefox launched in the guest will concurrently access the host profile, corrupting it. We can make it not do that by implementing locks, which would fix the corruption but would still break the browser in the guest. So let's explicitly set a muvm-private Firefox profile.

### DIFF
--- a/crates/muvm/src/env.rs
+++ b/crates/muvm/src/env.rs
@@ -90,6 +90,27 @@ pub fn prepare_env_vars(env: Vec<(String, Option<String>)>) -> Result<HashMap<St
     env_map.insert("GTK_IM_MODULE".to_owned(), "xim".to_owned());
     env_map.insert("QT_IM_MODULE".to_owned(), "xim".to_owned());
 
+    // Force a separate Firefox profile, since Firefox in muvm cannot see
+    // the lock from outside the VM and will concurrently launch on the same
+    // profile, corrupting it. This makes Firefox safely work in the VM
+    // (for browser launches).
+    if let Ok(home) = env::var("HOME") {
+        let mut path = PathBuf::new();
+        path.push(home);
+        path.push(".mozilla");
+        path.push("firefox");
+        if path.exists() {
+            path.push("muvm-profile");
+            if !path.exists() {
+                std::fs::create_dir(&path)?;
+            }
+            env_map.insert(
+                "XRE_PROFILE_PATH".to_owned(),
+                path.to_str().unwrap().to_owned(),
+            );
+        }
+    }
+
     for (key, value) in env {
         let value = value.map_or_else(
             || env::var(&key).with_context(|| format!("Failed to get `{key}` env var")),


teohhanhui:
Firefox communicates over d-bus to connect to an existing instance, right?

I think perhaps the right way to go is to forward that? Which IIUC will fix other things like some desktop integration too...

WhatAmISupposedToPutHere:
Dbus forwarding would be nice to have, but it is more complicated than it looks (there is fd passing involved), and you would need to have a "dbus firewall", as we will want certain objects to be separate between host and guest. I would say it is a bit too much rn, and the solution above works.

slp (approved):
LGTM

shaoyanji:
i was wondering about this. i run a nix home-manager firefox-esr to port my stuff over and each build corrupts the profile.ini file. my fix: to add a pre and posthook to backup and replace to keep the previous profile.ini to just keep the native firefox to work. curious if this works for the firefox-esr

asahilina:
That sounds like a different issue, I think? It's not profile.ini that gets corrupted, it's the actual profile databases (the ones that go quickly are cookies and places (history)). This happens because the VM does not share locks with the host (yet), so the normal locking that Firefox does on the profile does not work. If you are running multiple Firefox instances without a filesystem or OS layer that makes locks not work across them, they should not be able to launch concurrently on the same profile.

The environment variable probably works for all versions of Firefox though, if you want to try it.

Shados:
> and you would need to have a "dbus firewall", as we will want certain objects to be separate between host and guest

Worth noting that xdg-dbus-proxy does exist, motivated by similar goals within flatpak.

orowith2os:
> Worth noting that xdg-dbus-proxy does exist, motivated by similar goals within flatpak.

I would also like to mention that supporting dbus, and otherwise integrating better with other sockets and APIs from the host, would be very useful in running x86 Flatpak apps seamlessly under Apple Silicon.

With x86 apps needing 4KB pages, they'd crash immediately, even if you had a binfmt handler. You could fix this by using muvm, but without integrating with more than just GPU accel, portals and other APIs would fall apart immediately. Would be nice to look at exporting an x86 binfmt handler that runs inside of muvm; I know muvm already provides one, but I haven't checked if this is exposed to the host system at all.

waltmck:
> Dbus forwarding would be nice to have, but it is more complicated than it looks (there is fd passing involved), and you would need to have a "dbus firewall", as we will want certain objects to be separate between host and guest. I would say it is a bit too much rn, and the solution above works.

What objects would you want to be separate between the host and guest?

I know that some people use virtual machines to get isolation/containerization guarantees, but in my opinion these goals can be achieved more elegantly (most of the time) using native kernel features like `chroot` environments, network namespaces, capabilities, etc.

Speaking for myself, I really only need `muvm` in order to run FEX on a 16K machine. I don't need any containerization and I don't need any isolation---on the contrary, the lack of `dbus` services breaks many common apps that I would want to use (such as Slack). It would be great if there were a mode that aimed to provide maximum out-of-the-box integration with the host OS for those who don't care about isolation. A good start might be to pass environment variables and forward every socket in `/run`.

PR #121, opened 2024-12-08
muvm: remove RAM special case for 8GB machines

slp:
Now that muvm is able to request the guest to decrease its file cache and the root cause for the premature OOM has been indentified, let's remove the check for 8GB machines that limited the microVM to be smaller than 5GB, and just set it to be 80% of the host RAM, just as we do with systems with other configurations.

In practice, this enables 8GB machines to run more games out of the box (tested with multiple Steam games on a MacBook Air M1 with 8GB).


alyssarosenzweig (approved):
yay

asahilina (approved):

PR #122, opened 2024-12-09
Bump version to 0.2.0

slp:
Set up the stage for a new release. Bump minor version too, since this release includes breaking changes to the command line (-i behaves differently, and we default to muvm-x11bridge instead of sommelier).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,7 +428,7 @@ dependencies = [
 
 [[package]]
 name = "muvm"
-version = "0.1.4"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "bpaf",
--- a/crates/muvm/Cargo.toml
+++ b/crates/muvm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "muvm"
-version = "0.1.4"
+version = "0.2.0"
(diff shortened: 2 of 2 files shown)


slp:
If we merge #111 we won't be able to release an update on Fedora until we package all the dependencies, and that's going to take a while. I'd rather release this one now and then do another release in a couple of weeks when everything is ready.

slp:
Hm... we could rebase and merge #64 as an interim solution... WDYT?

WhatAmISupposedToPutHere:
If it is a couple weeks, i have no issues then, we can just carry #111 as a downstream patch in gentoo.

slp:
> If it is a couple weeks, i have no issues then, we can just carry #111 as a downstream patch in gentoo.

Out of curiosity, how does Gentoo deal with Rust dependencies? Does it just vendorize every crate?

n3ph:
@slp See https://github.com/chadmed/asahi-overlay/blob/main/app-emulation/muvm/muvm-0.1.4.ebuild#L6-L112. Crate versions are declared within ebuilds and downloaded/unpacked while emerge.

alyssarosenzweig (approved):

asahilina (approved):

PR #128, opened 2024-12-14
Support empty DISPLAY in xauth

ah-:
Currently x11bridge is broken on Gnome, because Gnome seems to create the wildcard entry in xauth with an empty DISPLAY. For example:
```~/t/muvm ❯❯❯ xauth list                                                                                                                                                                                          ✘ 1 
asahi/unix:  MIT-MAGIC-COOKIE-1  90e94fb5b89d6c1f5e512d58a1c6b8d3
#ffff#6173616869#:  MIT-MAGIC-COOKIE-1  90e94fb5b89d6c1f5e512d58a1c6b8d3
```

x11bridge then filters it out, resulting in an empty xauth file.

This makes x11bridge accept the Gnome entry, fixing X11.

I searched around a bit why Gnome is doing this, only thing I could find is https://xorg.freedesktop.narkive.com/zzRJsYVZ/is-xauth-entry-without-display-number-valid.

### DIFF
--- a/crates/muvm/src/guest/x11.rs
+++ b/crates/muvm/src/guest/x11.rs
@@ -65,7 +65,7 @@ where
             }
 
             // Check for the correct host display
-            if display != host_display.as_bytes() {
+            if !display.is_empty() && display != host_display.as_bytes() {
                 continue;
             }
 


slp:
Superseded by #129

teohhanhui:
This might be just a formality, but this PR was submitted 5 minutes before #129.

So personally I think we should rather say, I'm sorry that we missed this PR and stacked our approvals on the other one instead, even though the change was exactly the same.

slp:
Oops, you're right, @teohhanhui. Sorry @ah-, we should have reviewed and merged this one instead of #129.

PR #130, opened 2024-12-19
mount: Add support for FEX merged rootfs mode

asahilina:
In this mode, the FEX rootfs passed to FEX is already overlaid on top of the real root filesystem, so FEX directs all guest accesses to it instead of doing its own overlay logic. This, together with a bunch of fixes on the FEX side, fixes Wine.

Opt-in for now, since this actively *breaks* things with the current FEX. May become the default in the future once all that is sorted out.

The FEX config option name is speculative since it doesn't exist yet and it isn't strictly required to tell FEX about this, but I'd rather have it in to be able to change the logic on the FEX side in this mode as needed.

### DIFF
--- a/crates/muvm/src/guest/mount.rs
+++ b/crates/muvm/src/guest/mount.rs
@@ -1,15 +1,19 @@
+use std::collections::HashSet;
+use std::env;
 use std::ffi::CString;
-use std::fs::{read_dir, File};
+use std::fs::{read_dir, read_link, File};
 use std::io::Write;
 use std::os::fd::AsFd;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result};
 use rustix::fs::{mkdir, symlink, Mode, CWD};
 use rustix::mount::{
-    mount2, mount_bind, move_mount, open_tree, unmount, MountFlags, MoveMountFlags, OpenTreeFlags,
-    UnmountFlags,
+    mount2, mount_bind, mount_recursive_bind, move_mount, open_tree, unmount, MountFlags,
+    MoveMountFlags, OpenTreeFlags, UnmountFlags,
 };
+use rustix::path::Arg;
+use serde_json::json;
 
 fn make_tmpfs(dir: &str) -> Result<()> {
     mount2(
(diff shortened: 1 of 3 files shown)


asahilina:
Disabled the Clippy warning since, in this case, its suggestion actually makes the code flow less logical.

crates/muvm/src/guest/mount.rs
```diff
@@ -97,6 +221,24 @@ fn mount_fex_rootfs() -> Result<()> {
     Ok(())
 }
 
+pub fn overlay_file(src: &str, dest: &str) -> Result<()> {
+    let fd = open_tree(
+        CWD,
+        src,
+        OpenTreeFlags::OPEN_TREE_CLONE | OpenTreeFlags::OPEN_TREE_CLOEXEC,
+    )
+    .context("Failed to open_tree src")?;
```
teohhanhui, line 230:
```suggestion
    .with_context(|| format!("Failed to open_tree {src:?}"))?;
```

crates/muvm/src/guest/mount.rs
```diff
@@ -97,6 +221,24 @@ fn mount_fex_rootfs() -> Result<()> {
...
+
+    move_mount(
+        fd.as_fd(),
+        "",
+        CWD,
+        dest,
+        MoveMountFlags::MOVE_MOUNT_F_EMPTY_PATH,
+    )
+    .context("Failed to move_mount src to dest")
```
teohhanhui, line 239:
```suggestion
    .with_context(|| format!("Failed to move_mount {src:?} to {dest:?}"))
```

asahilina:
Also note that this depends on libkrun#244 (without it the overlayfs won't behave properly).

asahilina:
Added one unrelated cleanup commit because it depends on changes in the base commit here. Hopefully shouldn't be controversial.

crates/muvm/src/guest/mount.rs
```diff
@@ -60,33 +91,126 @@ fn mount_fex_rootfs() -> Result<()> {
...
     }
 
     // Now we need to tell FEX about this. One of the FEX share directories has an unmounted rootfs
     // and a Config.json telling FEX to use FUSE. Neither should be visible to the guest. Instead,
     // we want to replace the folders and tell FEX to use our mounted rootfs
     for base in ["/usr/share/fex-emu", "/usr/local/share/fex-emu"] {
         if Path::new(base).exists() {
-            let json = format!("{{\"Config\":{{\"RootFS\":\"{dir_rootfs}\"}}}}\n");
+            let json = if merged_rootfs {
+                format!("{{\"Config\":{{\"RootFS\":\"{dir_rootfs}\",\"MergedRootFS\":\"1\"}}}}\n")
```
teohhanhui, line 210: My eyes :see_no_evil:

crates/muvm/src/guest/mount.rs
```diff
@@ -60,33 +91,126 @@ fn mount_fex_rootfs() -> Result<()> {
...
+            } else {
+                format!("{{\"Config\":{{\"RootFS\":\"{dir_rootfs}\"}}}}\n")
+            };
```
teohhanhui, line 213:
```suggestion
            let json = if merged_rootfs {
                json!({
                    "Config": {
                        "RootFS": dir_rootfs,
                        "MergedRootFS": "1",
                    },
                })
            } else {
                json!({
                    "Config": {
                        "RootFS": dir_rootfs,
                    },
                })
            }.to_string();
```

https://docs.rs/serde_json/latest/serde_json/index.html#constructing-json-values
https://docs.rs/serde_json/latest/serde_json/macro.json.html
https://docs.rs/serde_json/latest/serde_json/enum.Value.html#impl-Display-for-Value

asahilina:
Reworked the logic a bit so merged_rootfs is ignored when the rootfs is host-managed (this means we can flip on the default without affecting those setups, which will make their own decision about merged rootfs behavior). Also unset the env var in that case, which should be cleaned up after !134 is merged since this should be moved to the config file.

slp (approved):
LGTM.

PR #132, opened 2024-12-20
mount: Symlink remaining FEX-Emu shared files

asahilina:
This is needed for thunks to work.

### DIFF
--- a/crates/muvm/src/guest/mount.rs
+++ b/crates/muvm/src/guest/mount.rs
@@ -88,8 +88,19 @@ fn mount_fex_rootfs() -> Result<()> {
         if Path::new(base).exists() {
             let json = format!("{{\"Config\":{{\"RootFS\":\"{dir_rootfs}\"}}}}\n");
             let path = base.to_string() + "/Config.json";
+            let host_dir = "/run/muvm-host".to_string() + base;
 
             make_tmpfs(base)?;
+            for entry in read_dir(host_dir).unwrap() {
+                let entry = entry.unwrap();
+                let file_name = entry.file_name();
+                if file_name == "Config.json" {
+                    continue;
+                }
+                let dest = Path::new(base).join(file_name);
+                symlink(entry.path(), dest)?;
+            }
+
             File::create(Path::new(&path))?.write_all(json.as_bytes())?;
         }
     }


alyssarosenzweig (approved):

alyssarosenzweig:
Possibly we want to invert the logic here and just overlay our Config.json over the dir, the way we handle /etc/resolv.conf currently? Not that it matters a ton either way

asahilina:
That doesn't work when Config.json doesn't already exist, I tried. The /etc/resolv.conf thing works because it does exist (as a broken symlink, which is enough).

alyssarosenzweig:
> That doesn't work when Config.json doesn't already exist, I tried. The /etc/resolv.conf thing works because it does exist (as a broken symlink, which is enough).

I see. Might be worth to put a config.json in the fedora fex pkg but that's neither here nor there. my approval stands.

crates/muvm/src/guest/mount.rs
```diff
@@ -88,8 +88,19 @@ fn mount_fex_rootfs() -> Result<()> {
         if Path::new(base).exists() {
             let json = format!("{{\"Config\":{{\"RootFS\":\"{dir_rootfs}\"}}}}\n");
             let path = base.to_string() + "/Config.json";
+            let host_dir = "/run/muvm-host".to_string() + base;
 
             make_tmpfs(base)?;
+            for entry in read_dir(host_dir).unwrap() {
+                let entry = entry.unwrap();
+                let file_name = entry.file_name().to_str().unwrap().to_owned();
+                if file_name == "Config.json" {
```
teohhanhui, line 97:
```suggestion
                let file_name = entry.file_name();
                if file_name == "Config.json" {
```

https://doc.rust-lang.org/std/ffi/struct.OsString.html#impl-Deref-for-OsString
https://doc.rust-lang.org/std/path/struct.Path.html#impl-PartialEq%3C%26OsStr%3E-for-Path
https://doc.rust-lang.org/std/path/struct.Path.html#method.join
https://doc.rust-lang.org/std/path/struct.Path.html#impl-AsRef%3CPath%3E-for-OsString

asahilina: Thanks! (This was a leftover from before I was using `Path` and doing direct string concat instead)

PR #134, opened 2024-12-26
Merge guest binaries.

WhatAmISupposedToPutHere:
Merge all guest binaries into one binary and process. This removes sommelier support, but people should use x11bridge anyway. Also stop leaking muvm configuration into vm processes.


crates/muvm/src/bin/muvm.rs
```diff
@@ -21,26 +22,27 @@ use muvm::launch::{launch_or_lock, LaunchResult, DYNAMIC_PORT_RANGE};
...
+use serde::Serialize;
 use tempfile::NamedTempFile;
 
-#[derive(Serialize, Deserialize)]
-pub struct KrunConfig {
+#[derive(Serialize)]
+pub struct KrunConfig<'a, 'b> {
     #[serde(rename = "Cmd")]
-    args: Vec<String>,
+    args: &'a [&'b str],
```
asahilina, line 38: I would leave this as `Vec<String>`. Lifetimes are kind of a premature optimization anti-pattern here, since they can make things more confusing. There's no issue copying a few strings around here.

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -13,12 +7,36 @@ use muvm::guest::server::server_main;
...
     env_logger::init();
 
-    let options = options().run();
+    let config_path = env::args()
+        .nth(1)
+        .context("expected configuration file path")?;
+    let mut config_file = File::open(&config_path)?;
+    let mut config_buf = Vec::new();
+    config_file.read_to_end(&mut config_buf)?;
+    _ = fs::remove_file(config_path);
```
asahilina, line 31: Why not propagate errors here?

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -13,12 +7,36 @@ use muvm::guest::server::server_main;
...
+    if let Ok(krun_config_path) = env::var(KRUN_CONFIG) {
+        _ = fs::remove_file(krun_config_path);
```
asahilina, line 33: Same as above, unless we expect this file to possibly not exist in some cases.

crates/muvm/src/bin/muvm.rs
```diff
@@ -407,16 +407,13 @@ fn main() -> Result<ExitCode> {
...
-        let display =
-            env::var("DISPLAY").context("X11 forwarding requested but DISPLAY is unset")?;
-        env.insert("HOST_DISPLAY".to_string(), display);
-
-        // And forward XAUTHORITY. This will be modified to fix the
-        // display name in muvm-guest.
-        if let Ok(xauthority) = env::var("XAUTHORITY") {
-            env.insert("XAUTHORITY".to_string(), xauthority);
-        }
+    let display = env::var("DISPLAY").context("X11 forwarding requested but DISPLAY is unset")?;
```
asahilina: `s/requested/required/`

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -80,13 +79,8 @@ fn main() -> Result<()> {
     let pulse_path = pulse_path.join("native");
     setup_socket_proxy(pulse_path, 3333)?;
 
-    if !setup_x11_forwarding(run_path)? {
-        // Will not return if successful.
-        exec_sommelier(&options.command, &options.command_args)
-            .context("Failed to execute sommelier")?;
-    }
+    setup_x11_forwarding(run_path)?;
```
asahilina: This should be changed to return `Result<()>` and change the false return case to an error.

crates/muvm/src/server/worker.rs
```diff
@@ -216,18 +218,42 @@ async fn handle_connection(
...
     if command == Path::new("/muvmdropcaches") {
-        let mut file = File::options()
-            .write(true)
-            .open("/proc/sys/vm/drop_caches")
-            .context("Failed to open /proc/sys/vm/drop_caches for writing")?;
-
-        {
-            file.write_all(b"1")
-                .context("Failed to write to /proc/sys/vm/drop_caches")?;
+        match unsafe { fork()? } {
```
asahilina, line 221: Needs SAFETY comment

crates/muvm/src/guest/x11.rs
```diff
@@ -89,6 +84,11 @@ where
         // See https://doc.rust-lang.org/std/env/fn.set_var.html#safety
         env::set_var("XAUTHORITY", dst_path);
     }
+    thread::spawn(|| {
+        if catch_unwind(|| start_x11bridge(1)).is_err() {
+            eprintln!("x11bridge thread crashed, x11 passthrough will no longer function");
```
asahilina, line 89: This is a bit annoying for debugging, since I often restarted x11bridge without restarting the whole VM, and also often wanted to strace it without stracing everything. Can we add a command line argument or something to just start x11bridge?

crates/muvm/src/guest/hidpipe.rs
```diff
@@ -119,7 +122,24 @@ fn init_uinput(sock: &mut UnixStream, user_id: u32) -> (u64, UInputHandle<File>)
...
-    chown(uinput.evdev_path().unwrap(), Some(user_id), Some(0)).unwrap();
+    let ev_path = CString::new(
+        uinput
+            .evdev_path()
+            .unwrap()
+            .into_os_string()
+            .into_encoded_bytes(),
+    )
+    .unwrap();
+    match unsafe { fork().unwrap() } {
```
asahilina, line 133: Since this is the same pattern as for drop_caches, maybe extract this into a `run_as_root()` function that takes a closure?

Also `SAFETY` comments needed.

crates/muvm/src/guest/x11.rs
```diff
@@ -9,14 +9,14 @@ use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
...
-pub fn setup_x11_forwarding<P>(run_path: P) -> Result<bool>
+pub fn setup_x11_forwarding<P>(run_path: P, host_display: Option<&str>) -> Result<bool>
 where
     P: AsRef<Path>,
 {
     // Set by muvm if DISPLAY was provided from the host.
-    let host_display = match env::var("HOST_DISPLAY") {
-        Ok(d) => d,
-        Err(_) => return Ok(false),
+    let host_display = match host_display {
```
asahilina, line 17: Per a previous comment, this is now required so it should not longer be an `Option<T>` at all.

asahilina:
FYI, I left a TODO in #130 about moving an env var to the muvm-guest config. If that gets merged first, please make that change too as part of the last commit here when you rebase.

crates/muvm/src/guest/x11.rs
WhatAmISupposedToPutHere: I would slightly prefer if we kept the command line parsing as simple as possible. Would a env variable be ok in this case?

crates/muvm/src/guest/bin/muvm-guest.rs
WhatAmISupposedToPutHere: Failing to remove this file is harmless, since it is on a tmpfs anyway. But i am not opposed to throwing an error here.

crates/muvm/src/guest/x11.rs
asahilina: Sure, it's just for debug anyway.

crates/muvm/src/guest/bin/muvm-guest.rs
asahilina: Just seems like better hygiene when the file is supposed to always exist, since we just read it ^^

teohhanhui: If it's not *expected* to fail, we could just `.expect(...)`

crates/muvm/src/guest/mount.rs
```diff
@@ -97,9 +92,6 @@ fn mount_fex_rootfs() -> Result<()> {
         // or is not installed at all. Avoid clobbering the config in that case.
         // merged_rootfs is ignored in this case, and we unset the env var so
```
asahilina, line 93: Drop everything after the comma (since we no longer use an env var).

asahilina:
Needs clippy fixes (probably just an `#[allow(clippy::too_many_arguments)]` in the problem functions).

crates/muvm/src/bin/muvm.rs
```diff
@@ -416,16 +416,13 @@ fn main() -> Result<ExitCode> {
     );
     env.insert("MUVM_SERVER_COOKIE".to_owned(), cookie.to_string());
 
-    if !options.sommelier {
```
sbrivio-rh, line 419: This breaks my (headless) usage:

```
$ ./target/debug/muvm -- bash
Error: X11 forwarding required but DISPLAY is unset

Caused by:
    environment variable not found
```
```
$ ./target/debug/muvm --sommelier -- bash
Error: X11 forwarding required but DISPLAY is unset

Caused by:
    environment variable not found
```

sbrivio-rh:
Ah, thanks, it works headless again. It's also a bit faster now. Tried with #111, without these commits:

```
$ time for i in `seq 1 10`; do ./target/debug/muvm --sommelier -- true; done

real	0m10.397s
user	0m7.653s
sys	0m3.165s
```

with:

```
$ time for i in `seq 1 10`; do ./target/debug/muvm -- true; done

real    0m10.100s
user    0m7.162s
sys     0m2.638s
```

By the way:

```
$ time for i in `seq 1 10`; do ./target/debug/muvm --passt-socket=/tmp/passt_3.socket -- true; done

real	0m9.808s
user	0m7.053s
sys	0m2.725s
```

...not really substantial.

sbrivio-rh:
Whee, it comes with `--privileged` too! \o/

WhatAmISupposedToPutHere:
It being a tiny bit faster is expected (exec is not free), but is not the reason for those changes.

crates/muvm/src/bin/muvm.rs
```diff
@@ -383,58 +362,59 @@ fn main() -> Result<ExitCode> {
     }
 
     let muvm_guest_path = find_muvm_exec("muvm-guest")?;
-    let muvm_server_path = find_muvm_exec("muvm-server")?;
 
-    let mut muvm_guest_args: Vec<String> = vec![
+    let display = env::var("DISPLAY").ok();
+    let guest_config = GuestConfiguration {
+        command: Launch {
```
teohhanhui, line 368: If I'm reading the code correctly, only `command` and `command_args` is actually used, right? Or is this intentional so that the user can pass in more stuff in the muvm config file in the future?

WhatAmISupposedToPutHere: Yes, that is correct, and somewhat intentional, i intend to make the first and the following commands no longer behave differently in the future

crates/muvm/src/bin/muvm.rs
```diff
@@ -383,58 +362,59 @@ fn main() -> Result<ExitCode> {
...
-        format!("{uid}", uid = getuid().as_raw()),
-        format!("{gid}", gid = getgid().as_raw()),
-        muvm_server_path
-            .to_str()
-            .context("Failed to process `muvm-server` path as it contains invalid UTF-8")?
-            .to_owned(),
-        command
-            .to_str()
-            .context("Failed to process command as it contains invalid UTF-8")?
             .to_string(),
```
teohhanhui, line 404:
```suggestion
            .to_owned(),
```

crates/muvm/src/bin/muvm.rs
```diff
@@ -383,58 +362,59 @@ fn main() -> Result<ExitCode> {
...
+        serde_json::to_string(&guest_config)
+            .context("Failed to transform GuestConfiguration into a JSON string")?
+    )
+    .context("Failed to write to temporary config file")?;
+
+    let muvm_config_path = muvm_config_file
+        .path()
+        .to_str()
+        .context("Temporary directory path contains invalid UTF-8")?
+        .to_string();
```
teohhanhui, line 399:
```suggestion
        .to_owned();
```

crates/muvm/src/bin/muvm.rs
```diff
@@ -383,58 +362,59 @@ fn main() -> Result<ExitCode> {
...
-
-        // And forward XAUTHORITY. This will be modified to fix the
-        // display name in muvm-guest.
-        if let Ok(xauthority) = env::var("XAUTHORITY") {
-            env.insert("XAUTHORITY".to_string(), xauthority);
-        }
+    // And forward XAUTHORITY. This will be modified to fix the
+    // display name in muvm-guest.
+    if let Ok(xauthority) = env::var("XAUTHORITY") {
+        env.insert("XAUTHORITY".to_string(), xauthority);
```
teohhanhui, line 411:
```suggestion
        env.insert("XAUTHORITY".to_owned(), xauthority);
```

crates/muvm/src/bin/muvm.rs
```diff
@@ -383,58 +362,59 @@ fn main() -> Result<ExitCode> {
...
     }
 
     let mut krun_config = KrunConfig {
```
teohhanhui, line 414:
```suggestion
    let krun_config = KrunBaseConfig {
        config: KrunConfig {
            args: muvm_guest_args,
            envs: env
                .into_iter()
                .map(|(key, value)| format!("{key}={value}"))
                .collect(),
        },
    };
```

(Sorry, GitHub's code suggestions UI is considerably bugged now...)

crates/muvm/src/bin/muvm.rs
```diff
@@ -383,68 +362,68 @@ fn main() -> Result<ExitCode> {
...
-    for (key, value) in env {
-        krun_config.envs.push(format!("{}={}", key, value));
-    }
     let krun_config = KrunBaseConfig {
-        config: krun_config,
+        config: KrunConfig {
+            args: muvm_guest_args,
+            envs: env
+                .into_iter()
+                .map(|(key, value)| format!("{}={}", key, value))
```
teohhanhui, line 419:
```suggestion
                .map(|(key, value)| format!("{key}={value}"))
```

No reason not to use the more readable syntax, since our MSRV is 1.77 >= 1.58

slp:
LGTM. I think we can merge this one once @teohhanhui suggestion is addressed.

slp:
@teohhanhui do you have any other concerns before merging this one?

teohhanhui (approved):
🚀

slp (approved):

PR #140, opened 2025-01-19
Use Box64 with muvm

m1m1k4tz:


crates/muvm/src/guest/box64.rs
```diff
@@ -0,0 +1,44 @@
+use std::fs::File;
+use std::io::Write;
+
+use anyhow::{Context, Result};
+
+use crate::utils::env::find_in_path;
+
+const BOX32_BINFMT_MISC_RULE: &str = ":FEX-x86:M:0:\\x7fELF\\x01\\x01\\x01\\x00\\x00\\x00\\x00\\\
```
asahilina, line 8: The rule names are still FEX

asahilina (changes requested):
Does box64 actually benefit from muvm? I thought they were doing the fake page size emulation thing...

This should probably be a command line argument, I'm not sure if automatically falling back to box64 makes sense. That could cause confusion for users who don't understand that the config is based on what packages are installed.

teohhanhui:
For some context:
* https://github.com/ptitSeb/box64/issues/1780#issuecomment-2577445931
* https://github.com/ptitSeb/box64/issues/1780#issuecomment-2577294869

I think we should close this PR, unless someone wants to take this on.

Otherwise we should just wait for @ptitSeb

ptitSeb:
> Does box64 actually benefit from muvm? 

Yes it does benifit from proper 4K pagesize

> I thought they were doing the fake page size emulation thing...

There is a fake 4k page size simulation, but it's not a perfect emulation. Far from it. It allow some linux program to be load easily, but everything that rely on 4k pagesize will just not work (like wine or steam).
 
> This should probably be a command line argument, I'm not sure if automatically falling back to box64 makes sense. That could cause confusion for users who don't understand that the config is based on what packages are installed.

While a command line argument is fine, I don't see why automatic fllback to box64 if present would be confusing.

m1m1k4tz:
> Does box64 actually benefit from muvm? I thought they were doing the fake page size emulation thing...


Right now muvm is required for wine to work under box64

alyssarosenzweig:
I'm OK with this

slp:
@m1m1k4tz @ptitSeb Do you mind if I pick up this one and improve it a bit in a new PR? I'd like to:

- Add a flag for explicitly choosing the emulator.
- Improve the commits title and summary.
- Add SoB.

ptitSeb:
> @m1m1k4tz @ptitSeb Do you mind if I pick up this one and improve it a bit in a new PR? I'd like to:
> 
> * Add a flag for explicitly choosing the emulator.
> * Improve the commits title and summary.
> * Add SoB.

I don't mind. Adding flags for explicitly choose one of the emu is good thing to add!

PR #143, opened 2025-01-31
Support using Box64 for x86 emulation

slp:
Add the "emu" command line argument to let users choose the emulator to be used for running x86 binaries. Add support for installing Box64 in binfmt_misc. If the argument is not present in the command line, try to use FEX first and if it can't be set up, try again with Box64.

Supersedes: #140

### DIFF
--- a/crates/muvm/src/cli_options.rs
+++ b/crates/muvm/src/cli_options.rs
@@ -5,6 +5,7 @@ use anyhow::{anyhow, Context};
 use bpaf::{any, construct, long, positional, OptionParser, Parser};
 
 use crate::types::MiB;
+use crate::utils::launch::Emulator;
 
 #[derive(Clone, Debug)]
 pub struct Options {
@@ -19,6 +20,7 @@ pub struct Options {
     pub tty: bool,
     pub privileged: bool,
     pub publish_ports: Vec<String>,
+    pub emulator: Option<Emulator>,
     pub command: PathBuf,
     pub command_args: Vec<String>,
 }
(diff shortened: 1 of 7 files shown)


slp:
cc/ @m1m1k4tz @ptitSeb

crates/muvm/src/cli_options.rs
```diff
@@ -60,6 +62,15 @@ pub fn options() -> OptionParser<Options> {
             None => Ok((s, None)),
         })
         .many();
+    let emulator = long("emu")
+        .help(
+            "Which emulator use for running x86_64 binaries.
+             Valid options are \"box\" and \"fex\". If this argument is not
+             present, muvm will try to use FEX, falling back to Box if it
+             it can't be found.",
```
teohhanhui, line 70:
```suggestion
            "Which emulator to use for running x86_64 binaries.
             Valid options are \"box\" and \"fex\". If this argument is not
             present, muvm will try to use FEX, falling back to Box if it
             can't be found.",
```

crates/muvm/src/guest/bin/muvm-guest.rs
```diff
@@ -77,7 +78,20 @@ fn main() -> Result<()> {
...
+            Emulator::Box => setup_box()?,
+            Emulator::Fex => setup_fex()?,
+        };
+    } else if let Err(err) = setup_fex() {
+        eprintln!("Error setting up FEX in binfmt_misc: {err}");
+        eprintln!("Failed to find or configure FEX, falling back to Box");
+
+        if let Err(err) = setup_box() {
+            eprintln!("Error setting up Box in binfmt_misc: {err}");
+            eprintln!("No emulators were configured, x86 emulation may nor work");
```
teohhanhui, line 92:
```suggestion
            eprintln!("No emulators were configured, x86 emulation may not work");
```

teohhanhui (approved):

ptitSeb:
looks good to me.

crates/muvm/src/cli_options.rs
slp: Fixed, thanks!

crates/muvm/src/guest/bin/muvm-guest.rs
slp: Fixed, thanks!

PR #147, opened 2025-02-20
krun-sys: find libkrun using pkg-config

slp:
This enables the use of non-system-wide installations of libkrun by exporting the right PKG_CONFIG_PATH.

### DIFF
--- a/crates/krun-sys/Cargo.toml
+++ b/crates/krun-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "krun-sys"
-version = "1.9.1"
+version = "1.10.1"
 edition = "2021"
 rust-version = "1.77.0"
 description = "Rust bindings for libkrun"
@@ -10,6 +10,7 @@ links = "krun"
 
 [build-dependencies]
 bindgen = { version = "0.69.4", default-features = false }
+pkg-config = { version = "0.3", default-features = false }
 
 [features]
 default = []
--- a/crates/krun-sys/build.rs
+++ b/crates/krun-sys/build.rs
@@ -1,13 +1,20 @@
 use std::env;
 use std::path::PathBuf;
 
-fn main() {
+fn main() -> Result<(), pkg_config::Error> {
     println!("cargo::rerun-if-changed=wrapper.h");
-    println!("cargo::rustc-link-lib=krun");
+
+    let library = pkg_config::probe_library("libkrun")?;
 
     let bindings = bindgen::Builder::default()
-        .header("wrapper.h")
+        .clang_args(
+            library
+                .include_paths
+                .iter()
+                .map(|path| format!("-I{}", path.to_string_lossy())),
+        )
         .clang_arg("-fretain-comments-from-system-headers")
+        .header("wrapper.h")
         .generate()
         .expect("Unable to generate bindings");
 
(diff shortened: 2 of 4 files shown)


teohhanhui:
We should investigate whether we can just use https://docs.rs/system-deps/latest/system_deps/

slp:
Wonderful, our CI uses Ubuntu, and `libkrun-dev` is not a package there. :-(

> We should investigate whether we can just use https://docs.rs/system-deps/latest/system_deps/
> 
> (as seem from [rust-lang/rust-project-goals#108 (comment)](https://github.com/rust-lang/rust-project-goals/issues/108#issuecomment-2502128887))

We're going to start using `krun-sys` on other projects, but we can move it to some other repo if `muvm` no longer needs it.

crates/krun-sys/build.rs
```diff
@@ -1,18 +1,29 @@
 use std::env;
 use std::path::PathBuf;
 
-fn main() {
+fn main() -> Result<(), pkg_config::Error> {
     println!("cargo::rerun-if-changed=wrapper.h");
-    println!("cargo::rustc-link-lib=krun");
```
teohhanhui, line 6: I'm most likely asking out of ignorance here, but https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib says this emits the `-l` flag to the compiler.

With this gone, where / how does that happen then? Below we only emit `-I` flags, right?

crates/krun-sys/build.rs
```diff
@@ -1,18 +1,29 @@
...
+
+    let library = pkg_config::probe_library("libkrun")?;
+
+    println!("cargo:warning={:?}", library.include_paths);
```
teohhanhui, line 9:
```suggestion
    println!("cargo::warning=include paths: {include_paths:?}", include_paths = library.include_paths);
```

(If this is meant to stay...)

crates/krun-sys/build.rs
slp: bindgen doesn't really need to link against the library, it just needs the headers.

crates/krun-sys/build.rs
slp: Oops, that shouldn't be there. I've just removed it, thanks!

crates/krun-sys/Cargo.toml
```diff
@@ -10,6 +10,7 @@ links = "krun"
 
 [build-dependencies]
 bindgen = { version = "0.69.4", default-features = false }
+pkg-config = { version = "0.3" }
```
teohhanhui, line 13: We should be explicit here and not use default features.

crates/krun-sys/build.rs
teohhanhui: Uhh... But since this crate is `krun-sys` (and specifically https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field), I think it's expected that it'll actually link to libkrun...

teohhanhui: But also, I'm confused where the linking to libkrun happens if not here?

crates/krun-sys/Cargo.toml
slp: Done, thanks.

crates/krun-sys/build.rs
slp: I was also getting very confused by this, because I was seeing the library getting injected at build time. Turns out `pkg_config::probe_library("libkrun")?;` also injects (by printing to stdout) parameters to the linker. The documentation of `pkg-config` is sparse, at best.

teohhanhui (approved):

PR #168, opened 2025-04-16
Return early from parsing DHCP response if End is encoutered

nrabulinski:
Up until now DHCP response was parsed by reading the option and the length first. The length field obviously does not apply to the End option, so if muvm got a response that ended exactly at the End byte, muvm would crash because of an out-of-bound read.

This works with older versions of passt because the response muvm received was longer than it could've been. Latest version of passt sends only as many bytes as necessary.

cc @sbrivio-rh

### DIFF
--- a/crates/muvm/src/guest/net.rs
+++ b/crates/muvm/src/guest/net.rs
@@ -171,6 +171,12 @@ fn do_dhcp(rtnl: &NlRouter) -> Result<()> {
 
         while p < len {
             let o = msg[p];
+
+            if o == 0xff {
+                // Option 255: End (of options)
+                break;
+            }
+
             let l: u8 = msg[p + 1];
             p += 2; // Length doesn't include code and length field itself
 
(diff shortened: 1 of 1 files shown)


sbrivio-rh:
> The length field obviously does not apply to the End option

Oops. It wasn't so obvious to me at some point, see: https://archives.passt.top/passt-dev/20250210191547.5cf0698a@elisabeth/, and I forgot to check against this implementation afterwards. :(

Thanks for fixing this!

`Reviewed-by: Stefano Brivio <sbrivio@redhat.com>`

teohhanhui (approved):

nrabulinski:
> Oops. It wasn't so obvious to me at some point

Haha yeah, hindsight is 20/20. Also, I did not realize you were behind passt (and the dhcp implementation in particular)! Thanks for your hard work! I should've paid more attention to commit authors 😁

sbrivio-rh:
You're welcome! I'm glad you use / like it!

PR #174, opened 2025-05-14
Set XDG_RUNTIME_DIR perimssions to 700

kitlith:
The current behavior seems to date back to the initial commit of the original C version of this tool:

https://github.com/teohhanhui/krun.legacy/commit/01ae759737f4056977c26e269978c670fd13e918#diff-ba1f55157a5e8661da39884a2669de7c912ff94d98134d6d520ed0468670e39bR198

Is anything relying on this extra permissiveness? I didn't notice any breakage in a quick test after patching this (consisting of launching Balatro through steam, nevermind that I normally just use the native build of love to run balatro) but that doesn't mean much.

### DIFF
--- a/crates/muvm/src/guest/user.rs
+++ b/crates/muvm/src/guest/user.rs
@@ -16,7 +16,7 @@ pub fn setup_user(uid: Uid, gid: Gid) -> Result<PathBuf> {
 
     let path = tempfile::Builder::new()
         .prefix(&format!("muvm-run-{uid}-"))
-        .permissions(Permissions::from_mode(0o755))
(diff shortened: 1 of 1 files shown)


teohhanhui:
Relevant section of the spec: 

> `$XDG_RUNTIME_DIR` defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored. The directory MUST be owned by the user, and they MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.

https://specifications.freedesktop.org/basedir-spec/latest/

teohhanhui (approved):

PR #179, opened 2025-06-07
Update nix to 0.30.1 to fix musl build

vimproved:
Currently muvm fails to build on musl due to getregs/setregs being missing from the nix crate on musl/aarch64. This was fixed in 0.30.0. As 0.30.0 also made unistd interfaces type-safe, this commit makes changes accordingly.


slp:
@vimproved thanks for the PR. Could you please rebase it to get a green light in the tests?

PR #189, opened 2025-09-01
Improved pipewire bridge support

WhatAmISupposedToPutHere:
Linux version of Dota 2 includes a new enough version of SDL to have native pipewire support. Naturally it ends up using protocol features that the bridge did not support previously


teohhanhui (approved):

slp:
LGTM, thanks! 

@WhatAmISupposedToPutHere should we publish a new release of muvm with this?

PR #190, opened 2025-09-02
Reduce bridge spamminess

WhatAmISupposedToPutHere:
Truncated messages are just a fact of life, we handle them correctly. Same for clients disconnecting with ECONNRESET, both were useful debug aids, but now they are just useless spam that can look like an error to an unknowing user.


slp:
@WhatAmISupposedToPutHere do you want to merge any more PRs before the release?

PR #193, opened 2025-09-17
Add a user init commands flag

valpackett:
Some custom init things (like spawning Wayland and D-Bus proxies) have to run **after** the user account has been set up and XDG_RUNTIME_DIR has been created and set. Let's simply add a big-X flag for executing things as the user?

### DIFF
--- a/crates/muvm/src/bin/muvm.rs
+++ b/crates/muvm/src/bin/muvm.rs
@@ -405,6 +405,7 @@ fn main() -> Result<ExitCode> {
         emulator: options.emulator,
         cwd,
         init_commands,
+        user_init_commands: options.user_init_commands,
     };
     let mut muvm_config_file = NamedTempFile::new()
         .context("Failed to create a temporary file to store the muvm guest config")?;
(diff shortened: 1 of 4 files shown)


slp (approved):
Sounds reasonable to me, thanks @valpackett

teohhanhui (approved):

teohhanhui:
@slp If we're okay with doing this over a semver major version, should we swap `-x` and `-X`? It seems much more intuitive.

WhatAmISupposedToPutHere (approved):

slp:
> @slp If we're okay with doing this over a semver major version, should we swap `-x` and `-X`? It seems much more intuitive.

The problem we have is that, for crates, bumping major is reasonably safe since cargo avoids picking it up unless instructed to do it. But, for binaries, I suspect people that may have integrated `muvm` into their scripts is likely not checking the version.

On the other hand, we don't know how prevalent is this use case, so perhaps we may as well do it once this time and see how many people complain?

PR #195, opened 2025-09-26
Add CLI flag --gpu=drm|venus|software to support various GPU virtualization modes

valpackett:
vDRM is great, but not supported everywhere yet, so on some machines the only way to get GPU acceleration is via Venus. Also, make it possible to use software rendering only while still having working cross-domain Wayland.

---

Should we also automatically set `MESA_LOADER_DRIVER_OVERRIDE=zink` when Venus is selected?


crates/muvm/src/bin/muvm.rs
```diff
@@ -185,7 +186,14 @@ fn main() -> Result<ExitCode> {
         }
 
         let virgl_flags = VIRGLRENDERER_USE_EGL
-            | VIRGLRENDERER_DRM
+            | VIRGLRENDERER_NO_VIRGL
```
teohhanhui, line 189: Is this flag supposed to be set unconditionally?

slp (approved):
LGTM, thanks!

PR #202, opened 2025-10-24
Add support for running a custom init (including systemd)

valpackett:
..yeah, you might say that this has really gone off the rails :) but we want to be able to ship apps with background services, configured the usual nix way just like on the host (which means systemd units are used to make them run).

Since https://github.com/containers/libkrun/pull/424 libkrun can run a custom init as PID 1, so now we can run systemd. It turned out to be quite easy to "disaggregate" muvm-guest into separate services that systemd can manage.. well, separately. Here I made it work like a busybox-style "mega binary" where what it does depends on what it's called as (`argv[0]`). I haven't yet added *all* the services, but I've already included pwbridge here as an example of supporting socket activation.

[muvm-systemd-pwb.av1.webm](https://github.com/user-attachments/assets/9bc63e65-2387-4f5d-a093-356845646b3f)

fun fact: I only found #201 because I went to implement this. :) and I did replicate that feature in systemd:

```ini
[Unit]
OnFailure=exit.target
OnSuccess=exit.target

[Service]
Type=exec
ExecStart=/opt/bin/muvm-remote /run/current-system/sw/bin/bash
ExecStopPost=+/nix/store/i0g30z11viax2v3ynm3p1411c4dhi6w8-python3-3.12.10-env/bin/python -c "import os,fcntl,struct;fcntl.ioctl(os.open('/', os.O_RDONLY), 0x7602, int(os.getenv('EXIT_STATUS', '1')))"
User=appvm
Group=appvm
StandardError=tty
StandardInput=tty
StandardOutput=tty
TTYPath=/dev/hvc0
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -501,6 +501,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "df1d3c3b53da64cf5760482273a98e575c651a67eec7f77df96b5b642de8f039"
 
 [[package]]
+name = "listenfd"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b87bc54a4629b4294d0b3ef041b64c40c611097a677d9dc07b2c67739fe39dba"
+dependencies = [
+ "libc",
+ "uuid",
+ "winapi",
+]
+
+[[package]]
 name = "lock_api"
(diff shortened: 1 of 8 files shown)


valpackett:
ping. any general feedback on this direction of work, on the changes currently on this branch?

PR #203, opened 2025-10-25
Create $XDG_RUNTIME_DIR in /run/user/{uid}

dramforever:
Using tempfile::Builder::tempdir() defaults to creating this directory in /tmp, which is shared with the host and often somewhat persistent. This causes one extra directory to be created in the host /tmp every time muvm runs.

Since we mount /run as a tmpfs now in the guest now, just create $XDG_RUNTIME_DIR in /run/user/{uid}, a common default. Specifically:

- Create /run/user as 0o755 (rwxr-xr-x) owned by root:root
- Create /run/user/{uid} as 0o700 (rwx------) owned by uid:gid

### DIFF
--- a/crates/muvm/src/guest/user.rs
+++ b/crates/muvm/src/guest/user.rs
@@ -1,9 +1,10 @@
 use std::env;
-use std::fs::{self, Permissions};
-use std::os::unix::fs::{chown, PermissionsExt as _};
+use std::fs;
+use std::os::unix::fs::chown;
 use std::path::{Path, PathBuf};
 
 use crate::guest::hidpipe::UINPUT_PATH;
+use crate::utils::fs::mkdir_mode;
 use anyhow::{anyhow, Context, Result};
 use nix::sys::wait::{waitpid, WaitStatus};
 use nix::unistd::{fork, setresgid, setresuid, ForkResult, Gid, Uid, User};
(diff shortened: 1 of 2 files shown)


dramforever:
I couldn't find a justification for using `/tmp`. This seems to have been there since the C krun's initial commit and might have just been missed, or maybe it was implemented before `/run` was properly handled.

dramforever:
AIUI the part where I changed is in the user initialization part, which happens after FEX init, so `guest/mount.rs` shouldn't be affected.

dramforever:
Oh, good idea. I'll change it to `/run/user/{uid}`

dramforever:
The directory and permissions now looks like this

```console
$ namei -l $XDG_RUNTIME_DIR 
f: /run/user/1000
drwxr-xr-x root root  /
drwxrwxrwt root root  run
drwxr-xr-x root root  user
drwx------ dram users 1000
```

slp:
> The other thing I could think of is, maybe this (`XDG_RUNTIME_DIR` in `/tmp`) was done this way to ease debugging? /cc @slp

IIRC it's in `/tmp` to avoid polluting the host's `/run` since originally we weren't mounting a `tmpfs` on it. I like the idea of switching to it now that we do the `tmpfs` mount.

dramforever:
Clippy should be happy now

PR #207, opened 2025-11-26
Add binfmt support files for x86_64 and i386

### LINKED ISSUE #24: Create binfmt support files
After #23 gets merged, krun will be capable of acting as a binfmt interpreter for `x86_64` binaries on aarch64 machines. We should create the binfmt support files, just qemu-user-static does, to package managers can install them on the host.

areofyl:
## Summary

This PR adds binfmt.d configuration files that allow package managers to register muvm as an interpreter for x86/x86_64 ELF binaries on aarch64 systems.

- Added `share/binfmt.d/muvm-x86_64.conf` for 64-bit x86 binaries
- Added `share/binfmt.d/muvm-i386.conf` for 32-bit x86 binaries

The configuration files follow the systemd binfmt.d format and use the same magic bytes and masks as qemu-user-static for compatibility. The `F` (fix-binary) flag is used to ensure the interpreter remains available in container and chroot environments.

Package maintainers can install these files to `/usr/lib/binfmt.d/` to enable transparent execution of x86 binaries through muvm.

Closes #24

## Test plan

- [ ] Verify configuration files are syntactically correct by registering them manually via `/proc/sys/fs/binfmt_misc/register`
- [ ] Test that x86_64 ELF binaries are correctly identified and passed to muvm
- [ ] Test that i386 ELF binaries are correctly identified and passed to muvm


WhatAmISupposedToPutHere:
Muvm is not yet capable of of being a binfmt_misc handler, so there is no point to adding configuration files at this point. Also, jfyi https://asahilinux.org/docs/project/policies/slop/

PR #220, opened 2026-04-15
muvm: add container mode and passt sandbox policy controls

yzewei:
## Background

When running `muvm` in a container environment, two common problems arise:

1. `hidpipe` depends on the host input device and is often unavailable within the container;

2. `passt` experiences sandbox degradation in containers with limited user namespaces, and its current behavior is not sufficiently controllable.

## Changes

We hope to add the following capabilities:

- Add `--no-hid`

- Disable HID device emulation paths, skipping hidpipe initialization.

- Added `--container-mode`

- Container-friendly default:

- Implicitly enable `--no-hid`

- If no passt policy is explicitly specified, default to `silent`

- Added `--passt-sandbox-policy=warn|strict|silent`

- `warn`: Preserve warnings and continue (default)

- `silent`: Suppress passt sandbox degradation warnings and continue

- `strict`: Require passt to enter an independent user namespace; otherwise, fail fast.

## Strict Policy Implementation Description

`strict` does not rely on log text for judgment, but directly checks the namespace:

- After starting passt, read `/proc/self/ns/user` and `/proc/<passt_pid>/ns/user`

- Poll for a maximum of 1 second

- If an independent user namespace is not entered, terminate passt and return an error.

## Compatibility

- Default behavior remains compatible (the original logic applies when no new parameters are used)

- `--container-mode` only takes effect when explicitly used

## Affected Files

- `crates/muvm/src/cli_options.rs`

- `crates/muvm/src/bin/muvm.rs`

- `crates/muvm/src/net.rs`

### DIFF
--- a/crates/muvm/src/bin/muvm.rs
+++ b/crates/muvm/src/bin/muvm.rs
@@ -15,7 +15,7 @@ use krun_sys::{
     VIRGLRENDERER_USE_ASYNC_FENCE_CB, VIRGLRENDERER_USE_EGL, VIRGLRENDERER_VENUS,
 };
 use log::debug;
-use muvm::cli_options::options;
+use muvm::cli_options::{options, PasstSandboxPolicy};
 use muvm::config::Configuration;
 use muvm::cpu::{get_fallback_cores, get_performance_cores};
 use muvm::env::{find_muvm_exec, prepare_env_vars};
@@ -75,6 +75,14 @@ fn main() -> Result<ExitCode> {
     }
 
     let options = options().fallback_to_usage().run();
+    let disable_hid = options.no_hid || options.container_mode;
+    let passt_policy = options
+        .passt_sandbox_policy
+        .unwrap_or(if options.container_mode {
+            PasstSandboxPolicy::Silent
+        } else {
+            PasstSandboxPolicy::Warn
+        });
     let config = Configuration::parse_config_file()?;
 
     let mut init_commands = options.init_commands;
@@ -278,7 +286,7 @@ fn main() -> Result<ExitCode> {
                 .context("Failed to connect to `passt`")?
                 .into()
         } else {
-            start_passt(&options.publish_ports)
+            start_passt(&options.publish_ports, passt_policy)
                 .context("Failed to start `passt`")?
                 .into()
         };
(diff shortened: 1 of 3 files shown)


yzewei:
@slp These alerts and "fail" errors are always bothering me when running in a container environment. I've considered restricting pass permissions and hidpip permissions within the container to solve this problem. Do you have any suggestions? PTAL!

teohhanhui:
LLM slop?

yzewei:
> LLM slop?


It's a real person, not AI. (Submitted).
If you have any objections, I can close this post.

WhatAmISupposedToPutHere:
Can you explain why do you want to actively disable hidpipe? if there are no input devices, it will just sit inert and not do anything.

yzewei:
> Can you explain why do you want to actively disable hidpipe? if there are no input devices, it will just sit inert and not do anything.

Currently, the `/dev/input` directory does not exist in many containers. `spawn_hidpipe_server()` will read `_dir("/dev/input")`, and if the directory does not exist, it will return an error, causing muvm to fail to start. In container scenarios, many workloads (CLI/services) typically do not require HID/raw input channels, and stdin/pty or graphical protocol inputs are more commonly used.

WhatAmISupposedToPutHere:
> > Can you explain why do you want to actively disable hidpipe? if there are no input devices, it will just sit inert and not do anything.
> 
> Currently, the `/dev/input` directory does not exist in many containers. `spawn_hidpipe_server()` will read `_dir("/dev/input")`, and if the directory does not exist, it will return an error, causing muvm to fail to start. In container scenarios, many workloads (CLI/services) typically do not require HID/raw input channels, and stdin/pty or graphical protocol inputs are more commonly used.

Well, in that cause can you make hidpipe cleanly exit if /dev/input is not present instead? that sounds like a better option.

yzewei:
> Well, in that cause can you make hidpipe cleanly exit if /dev/input is not present instead? that sounds like a better option.

I think this is a good idea, but do you have any good suggestions for the passt strategy? Of course, I could also choose not to add the container-mode and start the passt process on the host machine in advance, but it is more troublesome to use and does not really solve the problem of muvm running failure when passt fails to start from the sandbox.

WhatAmISupposedToPutHere:
> > Well, in that cause can you make hidpipe cleanly exit if /dev/input is not present instead? that sounds like a better option.
> 
> I think this is a good idea, but do you have any good suggestions for the passt strategy? Of course, I could also choose not to add the container-mode and instead relax the security policy restrictions when the container starts, but that's more complicated to use and doesn't really solve the problem of muvm failing to run when passt fails to start from the sandbox.

I have no idea what "sandbox degradation" even means, so no position on that one. Maybe you can explain the problem better?

yzewei:
> I have no idea what "sandbox degradation" even means, so no position on that one. Maybe you can explain the problem better

When muvm uses passt to establish network connections, it relies on passt. To ensure security, passt attempts to create an independent user namespace by calling `unshare(CLONE_NEWUSER)`.

However, nested user namespaces are prohibited in many container environments (due to seccomp).

Therefore, when it cannot create a user namespace (sandbox), passt will run in the current (container's) namespace and print an unsettling warning message (or exit directly depending on the configuration; if Seccomp/AppArmor is not disabled, passt will be killed by the kernel). 
For example:
```
muvm --container-mode --passt-sandbox-policy strict --cpu-list 0 /bin/echo OK
Error: Failed to start `passt`
Caused by:

0: `passt` sandbox policy check failed

1: `passt` exited before sandbox check completed with status exit status: 1
```

WhatAmISupposedToPutHere:
Sure. not sold on having an arg to control it, but making passt work makes sense.

yzewei:
> Sure. not sold on having an arg to control it, but making passt work makes sense.

Okay, I'll handle the input device not found issue separately first. This PR will be closed.

PR #221, opened 2026-04-16
muvm: gracefully disable hidpipe when /dev/input is missing

yzewei:
Treat missing /dev/input as non-fatal(container env), skip hidpipe setup, and avoid starting guest hidpipe when host hidpipe is disabled.


WhatAmISupposedToPutHere:
Fixed in #222, thank you.

PR #222, opened 2026-04-16
Do not assume /dev/input is present

WhatAmISupposedToPutHere:
Let's not make things complicated

Supersedes: #221

### DIFF
--- a/crates/muvm/src/hidpipe_server.rs
+++ b/crates/muvm/src/hidpipe_server.rs
@@ -283,9 +283,12 @@ where
 pub fn spawn_hidpipe_server(socket_path: PathBuf) -> anyhow::Result<()> {
     let mut evdevs = EvdevContainer::new();
     let epoll = Epoll::new(EpollCreateFlags::empty()).context("Failed to create epoll object")?;
-    for dir_ent in
-        fs::read_dir("/dev/input/").context("Failed to read \"/dev/input/\" directory")?
-    {
+    let dir_ents = match fs::read_dir("/dev/input/") {
+        Ok(dir_ents) => Some(dir_ents),
+        Err(e) if e.kind() == ErrorKind::NotFound => None,
+        Err(e) => return Err(e).context("Failed to read /dev/input/ directory"),
+    };
+    for dir_ent in dir_ents.into_iter().flatten() {
         let dir_ent = dir_ent.context("Failed to read directory entry")?;
         if dir_ent
             .file_type()


crates/muvm/src/hidpipe_server.rs
```diff
@@ -283,9 +283,12 @@ where
...
     let mut evdevs = EvdevContainer::new();
     let epoll = Epoll::new(EpollCreateFlags::empty()).context("Failed to create epoll object")?;
-    for dir_ent in
-        fs::read_dir("/dev/input/").context("Failed to read \"/dev/input/\" directory")?
-    {
+    let dir_ents = match fs::read_dir("/dev/input/") {
+        Ok(e) => e.collect(),
+        Err(e) if e.kind() == ErrorKind::NotFound => Vec::new(),
+        Err(e) => return Err(e).context("Failed to read \"/dev/input/\" directory"),
+    };
```
teohhanhui, line 290:
```suggestion
    let dir_ents = match fs::read_dir("/dev/input/") {
        Ok(dir_ents) => Some(dir_ents),
        Err(e) if e.kind() == ErrorKind::NotFound => None,
        Err(e) => return Err(e).context("Failed to read /dev/input/ directory"),
    };
    };
```

https://doc.rust-lang.org/std/option/enum.Option.html#impl-IntoIterator-for-Option%3CT%3E

teohhanhui (approved):
LGTM, just a nit

crates/muvm/src/hidpipe_server.rs
WhatAmISupposedToPutHere: This does not work, iterating over an iterator inside an option produces the iterator itself, not it's values.

WhatAmISupposedToPutHere: No, it is not: https://doc.rust-lang.org/src/core/option.rs.html#2601

Rustc disagrees too:
```
292 |         let dir_ent: () = dir_ent;
    |                      --   ^^^^^^^ expected `()`, found `ReadDir`

```
The `ReadDir` is the iterator type.

teohhanhui: Sorry. My bad. I missed another crucial part.

crates/muvm/src/hidpipe_server.rs
```diff
@@ -283,9 +283,12 @@ where
...
+    for dir_ent in dir_ents {
```
teohhanhui: This works. Did `cargo clippy` locally.

```suggestion
    let dir_ents = match fs::read_dir("/dev/input/") {
        Ok(dir_ents) => Some(dir_ents),
        Err(e) if e.kind() == ErrorKind::NotFound => None,
        Err(e) => return Err(e).context("Failed to read /dev/input/ directory"),
    };
    for dir_ent in dir_ents.into_iter().flatten() {
```

crates/muvm/src/hidpipe_server.rs
```diff
@@ -283,9 +283,12 @@ where
 pub fn spawn_hidpipe_server(socket_path: PathBuf) -> anyhow::Result<()> {
     let mut evdevs = EvdevContainer::new();
     let epoll = Epoll::new(EpollCreateFlags::empty()).context("Failed to create epoll object")?;
-    for dir_ent in
-        fs::read_dir("/dev/input/").context("Failed to read \"/dev/input/\" directory")?
-    {
+    let dir_ents = match fs::read_dir("/dev/input/") {
+        Ok(e) => Some(e),
```
teohhanhui: It's quite confusing to name this `e`, even if it's referring to directory **e**ntries. 😆

(And I love shadowing.)

```suggestion
        Ok(dir_ents) => Some(dir_ents),
```

PR #223, opened 2026-04-20
muvm-guest: pwbridge: forward fds on send

aford173:
Any outgoing Pipewire message with num_fd > 0 previously hit unimplemented!(), which in practice crashed the guest on the first ClientNode.PortBuffers that any pw-stream / pw-filter client emits.

Drain the expected fds from request_fds, convert each via vgpu_id_from_prime into a CrossDomainResource + GemHandleFinalizer, and attach them to the StreamSendResult. If request_fds is short of what the header claims, warn and return EIO instead of panicking.

Replace the unreachable!() PipeWireResourceFinalizer stub with an enum that dispatches to GemHandleFinalizer, so the finalizers produced here actually run.

Requires: https://github.com/containers/libkrun/pull/644

### DIFF
--- a/crates/muvm/src/guest/bridge/pipewire.rs
+++ b/crates/muvm/src/guest/bridge/pipewire.rs
@@ -4,15 +4,15 @@ use std::os::fd::{AsFd, AsRawFd, OwnedFd};
 use std::{env, mem};
 
 use anyhow::Result;
-use log::debug;
+use log::{debug, warn};
 use nix::errno::Errno;
 use nix::sys::epoll::EpollFlags;
 use nix::sys::eventfd::{EfdFlags, EventFd};
 
 use crate::guest::bridge::common;
 use crate::guest::bridge::common::{
-    Client, CrossDomainHeader, CrossDomainResource, MessageResourceFinalizer, ProtocolHandler,
-    StreamRecvResult, StreamSendResult,
+    Client, CrossDomainHeader, CrossDomainResource, GemHandleFinalizer, MessageResourceFinalizer,
+    ProtocolHandler, StreamRecvResult, StreamSendResult,
 };
 
 const CROSS_DOMAIN_CHANNEL_TYPE_PW: u32 = 0x10;
@@ -27,6 +27,7 @@ const PW_OPC_CORE_CREATE_OBJECT: u8 = 6;
 const PW_OPC_CORE_ADD_MEM: u8 = 6;
 const PW_OPC_CLIENT_UPDATE_PROPERTIES: u8 = 2;
 const PW_OPC_CLIENT_NODE_TRANSPORT: u8 = 0;
+const PW_OPC_CLIENT_NODE_PORT_BUFFERS: u8 = 5;
 const PW_OPC_CLIENT_NODE_SET_ACTIVATION: u8 = 10;
 
 #[repr(C)]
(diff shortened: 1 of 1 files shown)


WhatAmISupposedToPutHere:
Can you check which specific pipewire message is carrying those FDs? I do not want to unconditionally treat all fds as dma bufs, just in case there are things that need special handling.

crates/muvm/src/guest/bridge/pipewire.rs
```diff
@@ -349,13 +354,47 @@ impl ProtocolHandler for PipeWireProtocolHandler {
                     .insert(msg.new_id, ClientNodeData::new());
             }
         }
+        let mut resources = Vec::with_capacity(hdr.num_fd);
+        let mut finalizers = Vec::with_capacity(hdr.num_fd);
         if hdr.num_fd != 0 {
-            unimplemented!();
+            // ClientNode.PortBuffers is the only client->server method whose
```
WhatAmISupposedToPutHere: I think this belongs in commit description, or in docstring for the whole function.

aford173: The note is basically in the commit message, and a function-level docstring may be misleading since process_send_stream handles more than just fd forwarding, so I removed the comment.

WhatAmISupposedToPutHere (approved):

crates/muvm/src/guest/bridge/pipewire.rs
```diff
@@ -349,13 +354,39 @@ impl ProtocolHandler for PipeWireProtocolHandler {
...
         }
+        let mut resources = Vec::with_capacity(hdr.num_fd);
+        let mut finalizers = Vec::with_capacity(hdr.num_fd);
         if hdr.num_fd != 0 {
-            unimplemented!();
+            let is_port_buffers = hdr.opcode == PW_OPC_CLIENT_NODE_PORT_BUFFERS
+                && this.protocol_handler.client_nodes.contains_key(&hdr.id);
+            if !is_port_buffers {
+                warn!(
+                    "Pipewire send: unexpected fd-bearing msg id={} opc={} ({} fd[s])",
```
teohhanhui: We use [structured logging](https://docs.rs/log/latest/log/#structured-logging). So mostly there's no need to do formatting, unless the format of the message is actually important?

crates/muvm/src/guest/bridge/pipewire.rs
```diff
@@ -349,13 +354,39 @@ impl ProtocolHandler for PipeWireProtocolHandler {
...
+                    hdr.id, hdr.opcode, hdr.num_fd,
+                );
+                return Err(Errno::EIO.into());
+            }
+            if this.request_fds.len() < hdr.num_fd {
+                warn!(
+                    "Pipewire send: id={} opc={} expects {} fd(s), only {} queued",
```
teohhanhui: Same here.

crates/muvm/src/guest/bridge/pipewire.rs
```diff
@@ -165,13 +166,17 @@ impl PipeWireHeader {
...
+}
 
 impl MessageResourceFinalizer for PipeWireResourceFinalizer {
     type Handler = PipeWireProtocolHandler;
 
-    fn finalize(self, _: &mut Client<Self::Handler>) -> Result<()> {
-        unreachable!()
+    fn finalize(self, client: &mut Client<Self::Handler>) -> Result<()> {
+        match self {
+            PipeWireResourceFinalizer::Gem(fin) => fin.finalize(client),
```
teohhanhui:
```suggestion
            Self::Gem(fin) => fin.finalize(client),
```

nit

teohhanhui (changes requested):

teohhanhui (approved):

slp:
Good stuff here, thanks @aford173

PR #224, opened 2026-05-07
muvm: Add a --passt-args argument to pass args to passt

valpackett:
Sometimes it's necessary to pass additional args (such as the network interface) to passt. It's already been possible to spawn passt externally and give muvm its socket, but that's inconvenient especially when one wants to keep using the nicer muvm syntax for the exposed ports.

### DIFF
--- a/crates/muvm/Cargo.toml
+++ b/crates/muvm/Cargo.toml
@@ -24,6 +24,7 @@ procfs = { version = "0.18.0", default-features = false, features = [] }
 rustix = { version = "0.38.34", default-features = false, features = ["fs", "mount", "process", "pty", "std", "stdio", "system", "termios", "use-libc-auxv"] }
 serde = { version = "1.0.203", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.117", default-features = false, features = ["std"] }
+shell-words = { version = "1.1.1", default-features = false, features = ["std"] }
 tempfile = { version = "3.10.1", default-features = false, features = [] }
 tokio = { version = "1.38.0", default-features = false, features = ["io-util", "macros", "net", "process", "rt-multi-thread", "sync"] }
 tokio-stream = { version = "0.1.15", default-features = false, features = ["net", "sync"] }
--- a/crates/muvm/src/cli_options.rs
+++ b/crates/muvm/src/cli_options.rs
@@ -37,6 +37,7 @@ pub struct Options {
     pub mem: Option<MiB>,
     pub vram: Option<MiB>,
     pub passt_socket: Option<PathBuf>,
+    pub passt_args: Vec<String>,
     pub fex_images: Vec<String>,
     pub merged_rootfs: bool,
     pub interactive: bool,
@@ -135,6 +136,15 @@ pub fn options() -> OptionParser<Options> {
         .help("Instead of starting passt, connect to passt socket at PATH")
         .argument("PATH")
         .optional();
+    let passt_args = long("passt-args")
+        .help(
+            "When starting passt, append the given arguments.
+            May contain shell-quoted args, like so: --passt-args '-l \"my log file.txt\"'",
+        )
+        .argument::<String>("ARGS")
+        .parse(|s| shell_words::split(&s))
+        .many()
+        .map(|nested| nested.into_iter().flatten().collect());
     let interactive = long("interactive")
         .short('i')
         .help("Attach to the command's stdin/out after starting it")
@@ -192,6 +202,7 @@ pub fn options() -> OptionParser<Options> {
         mem,
         vram,
         passt_socket,
+        passt_args,
         fex_images,
         merged_rootfs,
         interactive,
--- a/crates/muvm/src/net.rs
+++ b/crates/muvm/src/net.rs
@@ -82,7 +82,7 @@ where
     Ok(UnixStream::connect(passt_socket_path)?)
 }
 
-pub fn start_passt(publish_ports: &[String]) -> Result<UnixStream> {
+pub fn start_passt(publish_ports: &[String], passt_args: &[String]) -> Result<UnixStream> {
     // SAFETY: The child process should not inherit the file descriptor of
     // `parent_socket`. There is no documented guarantee of this, but the
     // implementation as of writing atomically sets `SOCK_CLOEXEC`.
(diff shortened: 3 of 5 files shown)


crates/muvm/src/cli_options.rs
```diff
@@ -135,6 +136,10 @@ pub fn options() -> OptionParser<Options> {
         .help("Instead of starting passt, connect to passt socket at PATH")
         .argument("PATH")
         .optional();
+    let passt_args = long("passt-args")
+        .help("When starting passt, append the given arguments")
```
WhatAmISupposedToPutHere: Can you add an extended help message on how this interacts with shell word splitting

crates/muvm/src/net.rs
```diff
@@ -112,6 +112,13 @@ pub fn start_passt(publish_ports: &[String]) -> Result<UnixStream> {
     for spec in publish_ports {
         cmd.args(PublishSpec::parse(spec)?.to_args());
     }
+    for args in passt_args {
+        if let Ok(args) = shell_words::split(args) {
```
teohhanhui: I believe this is the wrong place to do this splitting. It should be done as part of the CLI options parsing? i.e. `passt_args` should be already in a form that's ready to be passed directly to `cmd.args(...)`

valpackett: sure

crates/muvm/src/cli_options.rs
```diff
@@ -37,6 +37,7 @@ pub struct Options {
     pub mem: Option<MiB>,
     pub vram: Option<MiB>,
     pub passt_socket: Option<PathBuf>,
+    pub passt_args: Vec<Vec<String>>,
```
teohhanhui: Is there any value in keeping the nested structure of the args, rather than flattening it while parsing? (This is a question.)

crates/muvm/src/cli_options.rs
WhatAmISupposedToPutHere: Sorry, still somewhat unclear as to what exactly "with shell quoting" means. Maybe add examples?

crates/muvm/Cargo.toml
```diff
@@ -24,6 +24,7 @@ procfs = { version = "0.18.0", default-features = false, features = [] }
 rustix = { version = "0.38.34", default-features = false, features = ["fs", "mount", "process", "pty", "std", "stdio", "system", "termios", "use-libc-auxv"] }
 serde = { version = "1.0.203", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.117", default-features = false, features = ["std"] }
+shell-words = { version = "1.1.1" }
```
teohhanhui:
```suggestion
shell-words = { version = "1.1.1", default-features = false, features = ["std"] }
```

WhatAmISupposedToPutHere:
Thinking of it, how bad would it be to instead do args "compiler driver"-style, so everything preceded by, say, -P gets sent to passt, and we do not do any shell expansion of our own? Eg: `-P --tcp-ports -P 22`

teohhanhui:
I think that would only make sense if we expect that passing passt-args would be a common usage?

WhatAmISupposedToPutHere:
> I think that would only make sense if we expect that passing passt-args would be a common usage?

Opposite, actually, this simplifies word splitting and does not need pulling in an extra crate in exchange for a slightly less convenient syntax

teohhanhui:
IMO short names are a very limited resource.

crates/muvm/src/cli_options.rs
valpackett: I just haven't realized that you even could do `.many()` first and then `.parse()` in bpaf.. but then if we do that, the management of the `Result` is becoming really annoying ><

valpackett:
Well we coooould do that style with a long arg, I guess xD

crates/muvm/src/cli_options.rs
teohhanhui: https://github.com/pacak/bpaf/blob/cc77231fea6cdbe182c5a1876df4d635b3952ca5/examples/many_comma_separated_args.rs

`collect` to the rescue again

valpackett: hm. More like the outer `map()` on the parser, and the number of intermediate `Vec` allocations is annoying but oh well

crates/muvm/src/cli_options.rs
```diff
@@ -135,6 +136,20 @@ pub fn options() -> OptionParser<Options> {
...
+            This option can be passed multiple times, and each string may contain multiple arguments,
+            each of which may be quoted according to the usual shell rules.",
+        )
+        .argument::<String>("ARGS")
+        .many()
+        .parse(|s| {
+            s.into_iter()
+                .map(|s| shell_words::split(&s))
+                .collect::<Result<Vec<_>, _>>()
+        })
```
teohhanhui: Actually, if I'm not mistaken, we actually want `.parse(...).many()` here instead of `.many().parse(...)`?

> Similarly, it is possible to parse multiple items with [many](https://docs.rs/bpaf/latest/bpaf/trait.Parser.html#method.many) or [some](https://docs.rs/bpaf/latest/bpaf/trait.Parser.html#method.some) by either parsing a single item first and then turning it into a [Vec](https://doc.rust-lang.org/nightly/alloc/vec/struct.Vec.html) or collecting them into a [Vec](https://doc.rust-lang.org/nightly/alloc/vec/struct.Vec.html) first and then parsing the whole vector. The former approach is more concise.

```suggestion
        .parse(|s| shell_words::split(&s))
        .many()
```

valpackett: that's where I started.. but right, it would also work with the flattening later!

crates/muvm/src/cli_options.rs
```diff
@@ -135,6 +136,20 @@ pub fn options() -> OptionParser<Options> {
         .help("Instead of starting passt, connect to passt socket at PATH")
         .argument("PATH")
         .optional();
+    let passt_args = long("passt-args")
+        .help(
+            "When starting passt, append the given arguments.
+            This option can be passed multiple times, and each string may contain multiple arguments,
+            each of which may be quoted according to the usual shell rules.",
```
teohhanhui: We're in "show, don't tell" territory here. 😆

So I'd suggest going back to a much simpler description, and just adding an example?

valpackett: We're in "too much bikeshedding" territory by now :laughing: Suure

teohhanhui (approved):

PR #229, opened 2026-06-05
Use the PULSE_SERVER env var to locate the pulseaudio socket

Tulon:
This change makes sound in muvm work in a Snap sandbox. The commit message explains exactly what the patch does and why it's necessary.

I've been applying this patch to muvm that I ship as part of the arm64 Snap build of [Wine Bar](https://github.com/Tulon/WineBar) for a month or two already - seems to work well.

Just keep in mind I don't really know Rust - the code is not necessarily the most optimal or elegant.

Also, I haven't tested it with the latest version of muvm. I currently ship `muvm-0.5.1` as part of WineBar.

### DIFF
--- a/crates/muvm/src/bin/muvm.rs
+++ b/crates/muvm/src/bin/muvm.rs
@@ -66,6 +66,38 @@ fn add_ro_disk(ctx_id: u32, label: &str, path: &str) -> Result<()> {
     }
 }
 
+fn pulse_path_from_xdg_runtime_dir(xdg_runtime_dir: &String) -> Result<CString> {
+    let pulse_path = Path::new(xdg_runtime_dir).join("pulse/native");
+    if pulse_path.exists() {
+        CString::new(
+            pulse_path
+                .to_str()
+                .expect("XDR_RUNTIME_DIR should not contain invalid UTF-8"),
+        )
+        .context("XDR_RUNTIME_DIR should not contain null characters")
+    } else {
+        Err(anyhow!(
+            "$XDR_RUNTIME_DIR/pulse/native (aka {}) doesn't exit",
+            pulse_path.to_string_lossy()
+        ))
+    }
+}
+
+fn pulse_path_from_pulse_server() -> Result<CString> {
+    let socket_uri = env::var("PULSE_SERVER")?;
+    let socket_path = socket_uri.strip_prefix("unix:").ok_or(anyhow!(
+        "Only the \"unix:\" scheme is supported in PULSE_SERVER"
+    ))?;
+
+    if Path::new(&socket_path).exists() {
+        CString::new(socket_path).context("PULSE_SERVER should not contain null characters")
+    } else {
+        Err(anyhow!(
+            "The socket in PULSE_SERVER ({socket_uri}) doesn't exit"
+        ))
+    }
+}
+
 fn main() -> Result<ExitCode> {
     env_logger::init();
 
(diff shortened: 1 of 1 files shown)


crates/muvm/src/bin/muvm.rs
```diff
@@ -291,23 +291,66 @@ fn main() -> Result<ExitCode> {
...
-                    .expect("pulse_path should not contain invalid UTF-8"),
-            )
-            .context("Failed to process `pulse/native` path as it contains NUL character")?;
-            // SAFETY: `pulse_path` is a pointer to a `CString` with long enough lifetime.
-            let err = unsafe { krun_add_vsock_port(ctx_id, PULSE_SOCKET, pulse_path.as_ptr()) };
-            if err < 0 {
-                let err = Errno::from_raw_os_error(-err);
-                return Err(err).context("Failed to configure vsock for pulse socket");
-            }
+    fn pulse_path_from_xdg_runtime_dir(xdg_runtime_dir: &Option<String>) -> Option<CString> {
```
WhatAmISupposedToPutHere: please no nested functions

crates/muvm/src/bin/muvm.rs
```diff
@@ -291,23 +291,66 @@ fn main() -> Result<ExitCode> {
...
+        match xdg_runtime_dir {
```
WhatAmISupposedToPutHere: this match can be a ?

crates/muvm/src/bin/muvm.rs
```diff
@@ -291,23 +291,66 @@ fn main() -> Result<ExitCode> {
...
+                if pulse_path.exists() {
+                    CString::new(
+                        pulse_path
+                            .to_str()
+                            .expect("XDR_RUNTIME_DIR should not contain invalid UTF-8"),
+                    )
+                    .context("XDR_RUNTIME_DIR should not contain null characters")
+                    .ok()
+                } else {
+                    eprintln!(
```
WhatAmISupposedToPutHere: is this needed?

crates/muvm/src/bin/muvm.rs
```diff
@@ -291,23 +291,66 @@ fn main() -> Result<ExitCode> {
...
+                        pulse_path.to_string_lossy()
+                    );
+                    None
+                }
+            },
+        }
+    }
+
+    fn pulse_path_from_pulse_server() -> Option<CString> {
+        match env::var("PULSE_SERVER") {
```
WhatAmISupposedToPutHere: this can be a ? with a .ok()

crates/muvm/src/bin/muvm.rs
```diff
@@ -291,23 +291,66 @@ fn main() -> Result<ExitCode> {
...
+            Err(_) => None,
+            Ok(socket_uri) => match socket_uri.strip_prefix("unix:") {
```
WhatAmISupposedToPutHere: this can be a ?

WhatAmISupposedToPutHere (changes requested):

crates/muvm/src/bin/muvm.rs
Tulon: Moved those outside of the parent function.

crates/muvm/src/bin/muvm.rs
Tulon: Sorry, I don't know rust in general and this particular syntax you are suggesting.

WhatAmISupposedToPutHere: `let pulse_path = Path::new(&xdg_runtime_path?).join("pulse/native");`

something like this

crates/muvm/src/bin/muvm.rs
Tulon: When I was debugging the issue of having no sound in muvm running in a Snap sandbox, there were no useful diagnostic messages as far as I remember. I decided to add some, just in case.

crates/muvm/src/bin/muvm.rs
Tulon: Again, not a rust guy.

crates/muvm/src/bin/muvm.rs
Tulon: Same as above.

crates/muvm/src/bin/muvm.rs
WhatAmISupposedToPutHere: Right, but you can return a Result here, and only do the error print if both of them fail

crates/muvm/src/bin/muvm.rs
Tulon: OK. Had to use `Path::new(xdg_runtime_dir.as_ref()?).join("pulse/native");` though.

crates/muvm/src/bin/muvm.rs
Tulon: I could, but then the error message would have less context. The way it's done now, not only does it tell you that the PA socket is missing but it also tells you where it got the path from.

Note that this won't be a spurious warning, as this function is called only as a last resort.

crates/muvm/src/bin/muvm.rs
Tulon: Implemented it as `.ok()?`.

crates/muvm/src/bin/muvm.rs
Tulon: I'd like to keep this `eprintln!("Only the \"unix:\" scheme is supported in PULSE_SERVER");` though.

I've now implemented it with `map_or_else(...)?` but if you ask me, that's way uglier than the original version with a `match`.

crates/muvm/src/bin/muvm.rs
```diff
@@ -66,6 +66,45 @@ fn add_ro_disk(ctx_id: u32, label: &str, path: &str) -> Result<()> {
...
+            "$XDR_RUNTIME_DIR/pulse/native (aka {}) doesn't exit",
+            pulse_path.to_string_lossy()
+        );
+        None
+    }
+}
+
+fn pulse_path_from_pulse_server() -> Option<CString> {
+    let socket_uri = env::var("PULSE_SERVER").ok()?;
+    let socket_path = socket_uri.strip_prefix("unix:").map_or_else(
```
WhatAmISupposedToPutHere: you can just use `.or_else` here

crates/muvm/src/bin/muvm.rs
WhatAmISupposedToPutHere: I still think that it should return a Result, for example your `.context` calls are useless, as the `.ok` immediately discards the error message.

crates/muvm/src/bin/muvm.rs
Tulon: Changed to `.or_else()`.

crates/muvm/src/bin/muvm.rs
Tulon: I've made it return `Result<CString>` though now the call site is way complex than before. Someone who doesn't know Rust really shouldn't be making such complex refactorings.

WhatAmISupposedToPutHere: The second helper also has the same issue

Tulon: Made the other one to also return `Result<CString>`.

WhatAmISupposedToPutHere (approved):

PR #231, opened 2026-06-29
x11bridge: tolerate redundant GEM close for multi-plane DRI3 buffers

aford173:
A DRI3 PixmapFromBuffers request for a multi-plane buffer (or any buffer whose planes share one backing object) passes several dma-buf fds that all resolve to the same virtgpu GEM handle, because drm_prime_fd_to_handle deduplicates imports of the same underlying object. vgpu_id_from_prime creates one GemHandleFinalizer per fd, so once the request has been forwarded the bridge calls drm_gem_close on that handle once per fd. The first close frees the handle; each subsequent close of the now-invalid handle returns EINVAL (ENOENT on some drivers), which propagated out of process_socket and tore down the entire X11 client connection.

This is easy to hit with the Intel iris native context, which hands DRI3 a multi-plane (CCS-aux) buffer: glmark2 and other GL clients died with "XIO: fatal IO error 95 ... after 38 requests" the moment the first frame's pixmap was finalized.

Treat EINVAL/ENOENT from the redundant close as success: the handle is already gone, which is exactly the intended outcome.

Tested on an Intel Raptor Lake-U iGPU (8086:a721) with the iris native context, with the host iGPU bound to both the i915 and the xe kernel driver: before this change every GL client died with the XIO 95 above; after it, glmark2 runs to completion on both drivers (GL_RENDERER "Mesa Intel(R) Graphics (RPL-U)", clean exit, no finalizer/XIO errors). The fix is host-driver-independent, as the offending close is against the guest virtio-gpu device.

### DIFF
--- a/crates/muvm/src/guest/bridge/common.rs
+++ b/crates/muvm/src/guest/bridge/common.rs
@@ -479,6 +479,9 @@ pub struct Client<'a, P: ProtocolHandler> {
     debug_loop: DebugLoop,
     pub send_queue: VecDeque<SendPacket>,
     pub sub_poll: SubPoll<'a, P>,
+    /// Live GemHandleFinalizers per GEM handle. Multi-plane / same-BO DRI3 buffers
+    /// dedup several fds to one handle, so we refcount and close only on the last.
+    gem_handle_refs: HashMap<u32, u32>,
 }
 
 #[derive(Debug)]
@@ -493,6 +496,16 @@ pub struct GemHandleFinalizer(u32);
 
 impl GemHandleFinalizer {
     pub fn finalize<T: ProtocolHandler>(self, client: &mut Client<T>) -> Result<()> {
+        // Several fds can dedup to one handle (multi-plane / same-BO buffers), so a
+        // finalizer per fd would close it more than once -- EINVAL/ENOENT, tearing
+        // down the client. Close only on the last reference; real errors then matter.
+        if let Some(refs) = client.gem_handle_refs.get_mut(&self.0) {
+            if *refs > 1 {
+                *refs -= 1;
+                return Ok(());
+            }
+            client.gem_handle_refs.remove(&self.0);
+        }
         // SAFETY: we own self.0
         unsafe {
             let close = DrmGemClose::new(self.0);
@@ -520,6 +533,7 @@ impl<'a, P: ProtocolHandler> Client<'a, P> {
             debug_loop: DebugLoop::new(),
             send_queue: VecDeque::new(),
             sub_poll,
+            gem_handle_refs: HashMap::new(),
         }));
         {
             let mut borrow = this.borrow_mut();
@@ -709,6 +723,11 @@ impl<'a, P: ProtocolHandler> Client<'a, P> {
         unsafe {
             drm_virtgpu_resource_info(self.gpu_ctx.fd.as_raw_fd() as c_int, &mut res_info)?;
         }
+        // Count a live reference; only the last finalizer for this handle closes it.
+        self.gem_handle_refs
+            .entry(to_handle.handle)
+            .and_modify(|refcount| *refcount += 1)
+            .or_insert(1);
         Ok((
             CrossDomainResource {
                 identifier: res_info.res_handle,


crates/muvm/src/guest/bridge/common.rs
```diff
@@ -494,11 +494,22 @@ pub struct GemHandleFinalizer(u32);
...
+            drm_gem_close(client.gpu_ctx.fd.as_raw_fd() as c_int, &close)
+        };
+        match res {
+            Ok(_) => Ok(()),
+            // A DRI3 PixmapFromBuffers for a multi-plane / same-BO buffer (as the
+            // Intel native context hands us) imports several fds that all resolve
+            // to the SAME GEM handle (drm_prime_fd_to_handle dedups). We create one
+            // finalizer per fd, so the handle gets closed more than once; the kernel
+            // returns EINVAL/ENOENT on the redundant close. The handle is already
+            // gone -- which is exactly what we wanted -- so treat that as success
```
teohhanhui: How do we tell an actual error apart from the case being described here? Could we keep track of fd's so that we know which ones are just redundant close calls (and just skip calling close altogether in that case)?

aford173: @teohhanhui  - I have an update coming with your suggestion to keep track of fd's.  I have tested it on Intel, and then I will test it on a Mediatek that doesn't appear to do multi-plane.

aford173: An update is pushed.  I also tested on AMD Strix Point as well as a Mediatek mt8196 to check for regressions on other hardware that didn't experience this before.

crates/muvm/src/guest/bridge/common.rs
```diff
@@ -479,6 +479,9 @@ pub struct Client<'a, P: ProtocolHandler> {
     debug_loop: DebugLoop,
     pub send_queue: VecDeque<SendPacket>,
     pub sub_poll: SubPoll<'a, P>,
+    // Live GemHandleFinalizers per GEM handle. Multi-plane / same-BO DRI3 buffers
+    // dedup several fds to one handle, so we refcount and close only on the last.
```
teohhanhui:
```suggestion
    /// Live GemHandleFinalizers per GEM handle. Multi-plane / same-BO DRI3 buffers
    /// dedup several fds to one handle, so we refcount and close only on the last.
```

Use doc comment here?

crates/muvm/src/guest/bridge/common.rs
```diff
@@ -493,12 +496,24 @@ pub struct GemHandleFinalizer(u32);
...
+        // down the client. Close only on the last reference; real errors then matter.
+        match client.gem_handle_refs.get_mut(&self.0) {
+            Some(refs) if *refs > 1 => {
+                *refs -= 1;
+                return Ok(());
+            }
+            _ => {
+                client.gem_handle_refs.remove(&self.0);
+            }
+        }
```
teohhanhui:
```suggestion
        if let Some(refs) = client.gem_handle_refs.get_mut(&self.0) {
            if *refs > 1 {
                *refs -= 1;
                return Ok(());
            }
            client.gem_handle_refs.remove(&self.0);
        }
```

crates/muvm/src/guest/bridge/common.rs
```diff
@@ -493,12 +496,24 @@ pub struct GemHandleFinalizer(u32);
...
         // SAFETY: we own self.0
-        unsafe {
+        let res = unsafe {
             let close = DrmGemClose::new(self.0);
-            drm_gem_close(client.gpu_ctx.fd.as_raw_fd() as c_int, &close)?;
-        }
-        Ok(())
+            drm_gem_close(client.gpu_ctx.fd.as_raw_fd() as c_int, &close)
+        };
+        res.map(|_| ()).map_err(Into::into)
```
teohhanhui: Revert all changes here?

crates/muvm/src/guest/bridge/common.rs
aford173: @teohhanhui  - Thanks for the quick reviews.  I think I have a 3rd attempt for your review.

crates/muvm/src/guest/bridge/common.rs
```diff
@@ -709,6 +723,8 @@ impl<'a, P: ProtocolHandler> Client<'a, P> {
         unsafe {
             drm_virtgpu_resource_info(self.gpu_ctx.fd.as_raw_fd() as c_int, &mut res_info)?;
         }
+        // Count a live reference; only the last finalizer for this handle closes it.
+        *self.gem_handle_refs.entry(to_handle.handle).or_insert(0) += 1;
```
teohhanhui:
```suggestion
        self.gem_handle_refs.entry(to_handle.handle).and_modify(|refcount| *refcount += 1).or_insert(1);
```

This seems to be more idiomatic:
* https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry
* https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.and_modify

teohhanhui:
CI failures seem unrelated. /cc @slp

@aford173 You need to fix the commit message to include the required `Signed-off-by: ...` signature.

aford173:
> CI failures seem unrelated. /cc @slp
> 
> @aford173 You need to fix the commit message to include the required `Signed-off-by: ...` signature.

Thanks for the reminder. Sorry about that.  I added it now.

teohhanhui (approved):

aford173:
@slp / @teohhanhui  - Any addition thoughts on if this could get merged?

WhatAmISupposedToPutHere (approved):

aford173:
To whom it may concern, I re-based on main, force-pushed, and the Code Quality test appears to pass now.

slp:
@aford173 Thanks for the rebase. Please sign the commit (git commit -s --amend) and we should be good to go.

aford173:
done

slp:
@aford173 Thanks!

PR #239, opened 2026-07-14
FEXInterpreter is now FEX, update your phonebooks

WhatAmISupposedToPutHere:


teohhanhui:
So we're okay with breaking compatibility with older versions of FEX, right?

teohhanhui (approved):

WhatAmISupposedToPutHere:
> So we're okay with breaking compatibility with older versions of FEX, right?

FEX has dropped the compatibility name, so i think we also can do it.

PR #243, opened 2026-09-17
host-guest: add an AT-SPI bridge over vsock

distantorigin:
Currently, assistive technologies cannot interact with or represent VM-bound applications.

Allow guest applications to use the host AT-SPI bus so assistive technologies on the host, such as Orca, can interact with them and receive accessibility events.

Use AT_SPI_BUS_ADDRESS when set; otherwise query org.a11y.Bus. If the address points to a reachable Unix socket, expose it through a dedicated vsock port. Skip the bridge when no supported socket is available.

muvm-guest provides a Unix socket under /run/user/{uid}, proxies connections to the host, and starts the bridge before user init commands. Set AT_SPI_BUS_ADDRESS, ATSPI_DISABLE_P2P, and GNOME_ACCESSIBILITY unless overridden with -e.

Rewrite EXTERNAL authentication to the VM owner identity and reject NEGOTIATE_UNIX_FD locally while preserving SASL reply order.

---

Please check all boxes that apply:

- [x] This code was developed in accordance with the Asahi Linux [Copyright Policy](https://asahilinux.org/copyright)
- [x] I have not used generative AI/LLMs to develop this code, in accordance with the Asahi Linux [LLM Policy](https://asahilinux.org/llm-policy)
- [x] I have certified the above by adding a `Signed-off-by:` trailer to *every* commit

All boxes must be checked before your pull request can be reviewed.


slp:
Nice PR here, thanks @distantorigin. Could you please fix the clippy issues and write down some brief instructions for testing this feature?

distantorigin:
Thanks for reviewing, happy to fix!

The clippy error seems to be caused by updates to Rust 1.98 and already existed in the repository before this pull request. Should I still fold that into this same PR?

https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#enhancements (#16746)

slp:
> The clippy error seems to be caused by updates to Rust 1.98 and already existed in the repository before this pull request. Should I still fold that into this same PR?

Ah, yes, you're right. I've created #245 to address this. We'll rebase on top of that to get a green CI.

distantorigin:
PR text updated to include testing steps, let me know if anything is unclear.

distantorigin:
@slp Rebased on the latest merge.

slp (approved):
LGTM, thanks @distantorigin !

